PR #8, opened 2023-05-18
Add placeholder templates dir

abays:
Adds a placeholder `templates` subdir and further content that will be useful once Barbican is fleshed-out


xek:
LGTM!

PR #10, opened 2023-05-22
Fix controller manager labeling

abays:
We now have a final form for operator controller-manager pod labeling of the format `openstack.org/operator-name: <operator name>` and are no longer relying on the inconsistent labels created by various `operator-sdk` versions.  All operators that lack this standardized pattern are being updated.


xek (commented):
LGTM

PR #14, opened 2023-07-06
Watch RBAC resources to trigger reconcile

kajinamit:
This ensures the controller watches service account, role, and role bindings to reconcile these resources in case any change is made.

### DIFF
--- a/controllers/barbican_controller.go
+++ b/controllers/barbican_controller.go
@@ -44,6 +44,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	"github.com/openstack-k8s-operators/lib-common/modules/database"
+	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


xek (approved):
LGTM

controllers/barbican_controller.go
```diff
@@ -282,6 +283,9 @@ func (r *BarbicanReconciler) reconcileDelete(ctx context.Context, instance *barb
 func (r *BarbicanReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&barbicanv1beta1.Barbican{}).
+		Owns(&corev1.ServiceAccount{}).
+		Owns(&rbacv1.Role{}).
+		Owns(&rbacv1.RoleBinding{}).
```
dmendiza, line 288: Hi, [kajinamit](https://github.com/kajinamit).  I was wondering why you're using `Owns` here instead of `Watches`.  I am still learning the framework, and if IIUC the Owns doc says these objects would be generated by the barbican operator?

> Owns defines types of Objects being *generated* by the ControllerManagedBy

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.0/pkg/builder#Builder.Owns

kajinamit: barbican controller DOES create these resources via ReconcileRbac .

dmendiza: I see, thanks.

PR #15, opened 2023-07-13
Fix defaulting of Replicas spec

kajinamit:
Currently the default values are ignored because these specs are set to 0 by webhook. Update the type from int32 to *int32 to avoid that behavior and ensure the expected default values are used.

This also removes the large replicas set in the sample config, because these are not suitable for testing.


dmendiza (approved):
LGTM

PR #16, opened 2023-07-13
Fix inconsistent password selector value for service password

kajinamit:
We use <Service Name>Password for service user passwords globally.


xek (approved):
LGTM

dmendiza (approved):
LGTM

PR #20, opened 2023-07-27
Add pre-commit config file

kajinamit:
... so that we can run the validation in CI later. This also fixes a few errors detected atm.


abays (approved):
/lgtm

PR #23, opened 2023-08-15
Fix `make run` errors

dmendiza:
This patch adds some missing hashes in `api/go.sum` that were causing `make run` to error out.


d34dh0r53 (approved):
LGTM

vakwetu (approved):

PR #25, opened 2023-08-17
Fix pre-commit failure

kajinamit:
This addresses the following error in pre-commit check.

check that scripts with shebangs are executable..........................Failed
- hook id: check-shebang-scripts-are-executable
- exit code: 1

templates/barbican/bin/init.sh: has a shebang but is not marked executable!
  If it is supposed to be executable, try: `chmod +x templates/barbican/bin/init.sh`
  If on Windows, you may also need to: `git add --chmod=+x templates/barbican/bin/init.sh`
  If it not supposed to be executable, double-check its shebang is wanted.


abays (approved):
/lgtm

PR #27, opened 2023-08-29
Add webhook support

abays:
Adds webhooks and also fixes a few permissions issues identified during the course of testing the former item.  Also fixes finalizers so that deletion works properly given what is currently deployed via the `Barbican` resource.


xek (approved):
Nice! I was just about to look into why finalizers don't work

abays:
Once this merges, using the operator locally will follow this process (as all other NextGen operators already do): https://github.com/openstack-k8s-operators/docs/blob/main/webhooks.md#using-webhooks-locally

vakwetu (approved):

PR #37, opened 2023-10-13
Shared github workflows

dprince:
This PR switches the operator to use shared build workflows for building operator, bundle, and catalog/index images.

Additionally align the opm version in the Makefile with other operators.

Jira: OSP-29146


vakwetu (approved):
LGTM

PR #38, opened 2023-10-19
Add override for barbican operator

vakwetu:
Untested as of yet. But this moves the creation of the route to the openstack operator so that we can take advantage of TLS-E enablement there.  This is a prereq for a patch to add the barbican operator to the openstack-operator


vakwetu:
This looks like its working correctly (at least locally).  I managed to test with kubefwd setting up the relevant paths for me.
We need this to merge so we can start integrating with openstack-operator

stuggi:
the change looks good to me

vakwetu:
acked by @dmendiza  , going to merge

PR #43, opened 2023-11-03
Fix operator installation

dmendiza:
This patch adds a missing CRD to kustomization.yaml that was causing the operator installation to fail.


xek (approved):
LGTM!

PR #44, opened 2023-11-06
Add missing RBAC config

dmendiza:
This patch adds the missing RBAC config for BarbicanWorker and BarbicanKeystoneListner that is causing the controller to get stuck in a crash loop.

### DIFF
--- a/config/rbac/barbicankeystonelistener_editor_role.yaml
+++ b/config/rbac/barbicankeystonelistener_editor_role.yaml
@@ -1,0 +1,31 @@
+# permissions for end users to edit barbicankeystonelisteners.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: clusterrole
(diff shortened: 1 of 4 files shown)


xek (changes requested):
I think these files are generated with make manifests. Are the proper kubebuilder annotations already there? If not, the pull request should update both places, like in this commit: https://github.com/openstack-k8s-operators/keystone-operator/commit/e6927b6989cfb40c65d5d82551620dabd38cb571

dmendiza:
Thanks for the review, @xek .  I had to do a bit of digging around in the operator sdk and kubebuilder docs to understand the annotations.  IIUC, the annotations are used by `make manifests` to generate the role.yaml file.  I added all the missing annotations to the Barbican controller and then updated the role.yaml file to use the one generated by the makefile.  The other files are generated by the openstack-sdk tool when a new api is added, and I think they were left out by mistake in the patch that implemented BarbicanKeystoneListener.

vakwetu (approved):
LGTM

xek (approved):

PR #45, opened 2023-11-09
Add more missing RBAC annotations

dmendiza:
This patch adds a couple more RBAC annotations that are needed to get barbican-operator running.


abays (approved):
/lgtm

PR #46, opened 2023-11-14
Barbican kuttl tests

mauricioharley:
Added KUTTL tests for the Barbican deployments.


xek (approved):
This looks fine, I'd say let's merge it, then try adding kuttl to the CI in the https://github.com/openstack-k8s-operators/release repo

test/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: barbican-worker-worker
+  ownerReferences:
+  - apiVersion: barbican.openstack.org/v1beta1
+    blockOwnerDeletion: true
+    controller: true
+    kind: BarbicanWorker
+    name: barbican-worker
```
vakwetu, line 51: Do we need these deployment objects to be defined here?  Isn't it sufficient to simply have the first stanza (lines 1-18)  I think we specifically say NOT to deploy the sub-components - which is why we add the top level component to the finalizers.

mauricioharley: @vakwetu 
The mentioned remaining lines can be removed from the test YAML files, but one of the tests scales all Barbican's components to zero. If we do not deploy them at the beginning of the tests, we need to pick one of the options below:

1. Deploy the components again after the "scale down to zero" test.
2. Remove the "scale down to zero" test.

Which approach should we choose?

vakwetu: I'm confused -- why wouldn't we have to redeploy the components after the scale to zero test?

Maybe thats precisely the point here - when you add an additional deployments, its not altogether clear what we are deploying.

As for the above question, there is option 3 - which is to make the scale to zero test the last test.  Then you won't have to deploy.  Or you can do a scale up test after the scale down test.

mauricioharley: The "scale down to zero" test is already the last one. It makes sense to be like this.

On the local environment I'm using to test the tests, they are running on the same namespace where the other CRs are deployed (Memcached, Redis, RabbitMQ, ...) since to be able to work, they need such components to be present at first.

I said we need to manually redeploy the components simply because the "scale down to zero" test deletes them. May I understand we won't need the CRs after they are deleted, considering the namespace statement above? I think a quick sync on this would be better than exchanging comments here. :-)

vakwetu: As we discussed, I realized while we chatted that this is an assert-file.  So everything I said about removing the deployments is wrong.

vakwetu: So keep them, but also make sure to test using install_yamls.

PR #47, opened 2023-11-14
Remove duplicated ContainerImage

dmendiza:
This patch removes the ContainerImage property from BarbicanTemplate to avoid a name collision with BarbicanComponentTemplate in components that inline both templates.


vakwetu (approved):
/LGTM

PR #48, opened 2023-11-16
Add CRDs under api

vakwetu:
This is something that has changed in the Makefile since we started. Importantly, the functional tests in openstack-operator and possibly other places expect this to be there so that the crds can be loaded.


abays (approved):
/lgtm

xek (approved):

PR #49, opened 2023-11-20
Properly set certain conditions to True when necessary

abays:
This fixes the conditions for the various `Barbican*` CRDs such that the top-level `Barbican` CR can reach the "ready" state.  By fixing this, we also allow Barbican to properly integrate with the OpenStack operator, as the latter operator relies on the `Barbican` CR reaching the ready state in order for its own `OpenStackControlPlane` CR to reach the ready state.

### DIFF
--- a/controllers/barbican_controller.go
+++ b/controllers/barbican_controller.go
@@ -164,7 +164,6 @@ func (r *BarbicanReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 			condition.UnknownCondition(barbicanv1beta1.BarbicanAPIReadyCondition, condition.InitReason, barbicanv1beta1.BarbicanAPIReadyInitMessage),
 			condition.UnknownCondition(barbicanv1beta1.BarbicanWorkerReadyCondition, condition.InitReason, barbicanv1beta1.BarbicanWorkerReadyInitMessage),
 			condition.UnknownCondition(barbicanv1beta1.BarbicanKeystoneListenerReadyCondition, condition.InitReason, barbicanv1beta1.BarbicanKeystoneListenerReadyInitMessage),
-			condition.UnknownCondition(condition.DeploymentReadyCondition, condition.InitReason, condition.DeploymentReadyInitMessage),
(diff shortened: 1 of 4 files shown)


controllers/barbicanapi_controller.go
```diff
@@ -140,7 +140,6 @@ func (r *BarbicanAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			condition.UnknownCondition(condition.ServiceConfigReadyCondition, condition.InitReason, condition.ServiceConfigReadyInitMessage),
 			condition.UnknownCondition(condition.DeploymentReadyCondition, condition.InitReason, condition.DeploymentReadyInitMessage),
 			// right now we have no dedicated KeystoneServiceReadyInitMessage and KeystoneEndpointReadyInitMessage
-			condition.UnknownCondition(condition.KeystoneServiceReadyCondition, condition.InitReason, ""),
```
abays, line 143: This was not used anywhere, so I removed it.

controllers/barbican_controller.go
```diff
@@ -164,7 +164,6 @@ func (r *BarbicanReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 			condition.UnknownCondition(barbicanv1beta1.BarbicanAPIReadyCondition, condition.InitReason, barbicanv1beta1.BarbicanAPIReadyInitMessage),
 			condition.UnknownCondition(barbicanv1beta1.BarbicanWorkerReadyCondition, condition.InitReason, barbicanv1beta1.BarbicanWorkerReadyInitMessage),
 			condition.UnknownCondition(barbicanv1beta1.BarbicanKeystoneListenerReadyCondition, condition.InitReason, barbicanv1beta1.BarbicanKeystoneListenerReadyInitMessage),
-			condition.UnknownCondition(condition.DeploymentReadyCondition, condition.InitReason, condition.DeploymentReadyInitMessage),
```
abays, line 167: This was not used anywhere, so it was removed.  It may have been a copy-paste from something like the Cinder operator, which also has it but does not use it.  In the Cinder case though, it is not considered in calculating the top-level CR's overall "ready" status, so it does not interfere (but I am creating a PR to remove it there as well).

vakwetu (approved):
/lgtm

dmendiza (approved):

PR #53, opened 2023-12-08
Make Secure RBAC configurable

dmendiza:
This patch adds a new field to the BarbicanAPI spec to configure "Consistent and Secure RBAC" [1].  The API configuration template was modified to use the value in this field instead of being hard-coded to always be on.

Jira: OSPRH-2129

[1] https://governance.openstack.org/tc/goals/selected/consistent-and-secure-rbac.html

### DIFF
--- a/controllers/barbican_controller.go
+++ b/controllers/barbican_controller.go
@@ -539,12 +539,13 @@ func (r *BarbicanReconciler) generateServiceConfig(
 			instance.Status.DatabaseHostname,
 			barbican.DatabaseName,
 		),
-		"KeystoneAuthURL": keystoneInternalURL,
-		"ServicePassword": string(ospSecret.Data[instance.Spec.PasswordSelectors.Service]),
-		"ServiceUser":     instance.Spec.ServiceUser,
-		"ServiceURL":      "TODO",
-		"TransportURL":    string(transportURLSecret.Data["transport_url"]),
-		"LogFile":         fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
+		"KeystoneAuthURL":  keystoneInternalURL,
+		"ServicePassword":  string(ospSecret.Data[instance.Spec.PasswordSelectors.Service]),
+		"ServiceUser":      instance.Spec.ServiceUser,
+		"ServiceURL":       "TODO",
+		"TransportURL":     string(transportURLSecret.Data["transport_url"]),
+		"LogFile":          fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
+		"EnableSecureRBAC": instance.Spec.BarbicanAPI.EnableSecureRBAC,
 	}
 
 	return GenerateConfigsGeneric(ctx, h, instance, envVars, templateParameters, customData, labels, false)
--- a/controllers/barbicanapi_controller.go
+++ b/controllers/barbicanapi_controller.go
@@ -279,12 +279,13 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
 			instance.Spec.DatabaseHostname,
 			barbican.DatabaseName,
 		),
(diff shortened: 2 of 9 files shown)


controllers/barbican_controller.go
```diff
@@ -539,12 +539,13 @@ func (r *BarbicanReconciler) generateServiceConfig(
...
-		"ServiceURL":      "TODO",
-		"TransportURL":    string(transportURLSecret.Data["transport_url"]),
-		"LogFile":         fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
+		"KeystoneAuthURL":  keystoneInternalURL,
+		"ServicePassword":  string(ospSecret.Data[instance.Spec.PasswordSelectors.Service]),
+		"ServiceUser":      instance.Spec.ServiceUser,
+		"ServiceURL":       "TODO",
+		"TransportURL":     string(transportURLSecret.Data["transport_url"]),
+		"LogFile":          fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
+		"enableSecureRBAC": instance.Spec.BarbicanAPI.EnableSecureRBAC,
```
vakwetu, line 548: nit: We should be consistent here and use "EnableSecureRBAC" instead  of "enableSecureRBAC"

controllers/barbicanapi_controller.go
```diff
@@ -279,12 +279,13 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
-		"ServiceURL":      "https://barbican.openstack.svc:9311",
-		"TransportURL":    string(transportURLSecret.Data["transport_url"]),
-		"LogFile":         fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
+		"KeystoneAuthURL":  keystoneInternalURL,
+		"ServicePassword":  string(ospSecret.Data[instance.Spec.PasswordSelectors.Service]),
+		"ServiceUser":      instance.Spec.ServiceUser,
+		"ServiceURL":       "https://barbican.openstack.svc:9311",
+		"TransportURL":     string(transportURLSecret.Data["transport_url"]),
+		"LogFile":          fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
+		"enableSecureRBAC": instance.Spec.EnableSecureRBAC,
```
vakwetu, line 288: ditto

templates/barbican/config/00-default.conf
```diff
@@ -28,8 +28,8 @@ topic = barbican_notifications
 driver=messagingv2
 
 [oslo_policy]
-enforce_scope = true
-enforce_new_defaults = true
+enforce_scope = {{ .enableSecureRBAC }}
```
vakwetu, line 31: Would we ever have a situation where we would want enforce_scope and enforce_new_defaults to be different from each other?

vakwetu (changes requested):
Just a small nit on consistency of variable names.  Lets fix that before we merge.
Otherwise, looks pretty straightforward.

templates/barbican/config/00-default.conf
dmendiza: I don't think so. We've only ever tested both off or both on.  I guess it's possible to do, but not something you'd want.

e.g. scope off and new defaults on would allow anyone with a reader role to access APIs intended for operators.  OTOH, scope on and new defaults off would enforce scope in the legacy policies (which are not written with scope in mind) so you'd probably get some weird behaviors.

controllers/barbican_controller.go
dmendiza: fixed

controllers/barbicanapi_controller.go
dmendiza: fixed

vakwetu (approved):
/lgtm

PR #56, opened 2023-12-15
Rename KUTTL tests dir to fix the tests on install_yamls

mauricioharley:
The KUTTL tests are currently running, but install_yamls considers them to be in the "tests" directory, rather than "test", as it currently is.


abays (approved):
/lgtm

PR #57, opened 2023-12-18
Add missing net-attach-def permissions

abays:
This is needed in case the the user includes any `networkAttachments` in the various CRs.  Otherwise we see this:

```
E1218 13:13:53.990846       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.9/tools/cache/reflector.go:169: Failed to watch *v1.NetworkAttachmentDefinition: failed to list *v1.NetworkAttachmentDefinition: network-attachment-definitions.k8s.cni.cncf.io is forbidden: User "system:serviceaccount:openstack-operators:barbican-operator-controller-manager" cannot list resource "network-attachment-definitions" in API group "k8s.cni.cncf.io" at the cluster scope
```

### DIFF
--- a/controllers/barbican_controller.go
+++ b/controllers/barbican_controller.go
@@ -89,6 +89,8 @@ func (r *BarbicanReconciler) GetLogger(ctx context.Context) logr.Logger {
 //+kubebuilder:rbac:groups=rabbitmq.openstack.org,resources=transporturls,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbdatabases,verbs=get;list;watch;create;update;patch;delete;
+//+kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
+
 // service account, role, rolebinding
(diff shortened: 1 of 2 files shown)


controllers/barbican_controller.go
```diff
@@ -89,6 +89,8 @@ func (r *BarbicanReconciler) GetLogger(ctx context.Context) logr.Logger {
 //+kubebuilder:rbac:groups=rabbitmq.openstack.org,resources=transporturls,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbdatabases,verbs=get;list;watch;create;update;patch;delete;
+//+kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
```
mauricioharley, line 92: Don't we need the "delete" or "update" verbs as well?

abays: No, because we're not actually creating/updating/deleting the NAD resources.  We're only checking that they exist when someone specifies an attachment in the CR specs, i.e.

```
    networkAttachments:
      - storage
```

The NADs themselves are created manually by the user beforehand i.e. https://github.com/openstack-k8s-operators/architecture/blob/main/va/hci/stage3/netattach_storage.yaml

xek (approved):

PR #74, opened 2024-01-12
Use Simple Crypto KEK from Secret

dmendiza:
This patch changes the configuration template to use a KEK provided in the OSP secret instead of a hard-coded one.


vakwetu (changes requested):
You've added the KEK to the barbican_api, but doesn't either the keystone-listener (probably not), barbican-worker (probably) or dbsync job (probably not?) need this too?

vakwetu (approved):

PR #77, opened 2024-01-16
Fix intermittent connection refused

fmount:
During the test of `Glance` image sign that rely on `Barbican` to store secrets, we’d see `Connection refused` errors on about 60% of the calls made to the service. This let the image signing fail most of the times.
In addition, simply running `curl` against the `barbican-internal` endpoint shows this issue, because it fails more than **half** of the times with a `connection refused` error on 9311.
However, if we point directly to the `Pod` (bypassing the k8s `Service`), we can see that this problem doesn't occur, and we're able to always reach the `Barbican` service and successfully sign the `Glance` images.
The root cause of this issue is the k8s `Service` sending traffic to the `BarbicanAPI`, as well as `keystoneListener` and `BarbicanWorker`: these two services are not listening on `9311`, and this result in a connection refused on that port.
The `BarbicanAPI` service should be built with a selector that is able to forward the traffic to `barbicanAPI` only, and this patch improves the way labels are used to achieve this goal, similarly to what we did in `Cinder`, `Glance` and `Manila`.
With this patch we no longer see any issue when we try to reach the `barbicanAPI` component and we're able to retrieve the secrets and sign images.

Fixes: [OSPRH-3420](https://issues.redhat.com//browse/OSPRH-3420)

### DIFF
--- a/controllers/barbicanapi_controller.go
+++ b/controllers/barbicanapi_controller.go
@@ -574,7 +574,8 @@ func (r *BarbicanAPIReconciler) reconcileNormal(ctx context.Context, instance *b
 
 	Log.Info(fmt.Sprintf("[API] Getting service labels '%s'", instance.Name))
 	serviceLabels := map[string]string{
-		common.AppSelector: fmt.Sprintf(barbican.ServiceName),
+		common.AppSelector:       fmt.Sprintf(barbican.ServiceName),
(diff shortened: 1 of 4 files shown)


konan-abhi (approved):
Looks good!!
Thank you!

abays (approved):
/lgtm

PR #111, opened 2024-04-01
Re-init conditions each reconcile

fmount:
This patch follows the same pattern applied to the other operators, where we re-init the condition at each reconcile loop. `Condition`s are re-evaluated and updated, keeping the `LastTransitionTime` for those that haven't changed (it avoids the transition from `True` to `Unknown` to `True` again).
In addition, the `observedGeneration` field is introduced, and it is used by the `openstack-operator` to check the `IsReady()` function for a particular `CR` in case a minor update is triggered. All the conditions are evaluated during the main `Reconcile` loop ( or the `reconcileNormal` function in some circumstances), hence the main `ReadyCondition` is updated within the same flow. The `defer` function still updates the `Resource` and `Mirror(s)` the condition to the top-level `CR`.

Jira: [OSPRH-5919](https://issues.redhat.com//browse/OSPRH-5919)
Jira: [OSPRH-5698](https://issues.redhat.com/browse/OSPRH-5698)


abays (approved):
/lgtm

PR #113, opened 2024-04-23
Ensure golangci-lint runs on all files

gibizer:
* removed unused funcs, func params, constants, and imports
* ignore revive dot-imports rule on ginkgo and gomega as dot import
there is the recommended practice
* various small style fixes found by the linter
* removed unused suite_test.go from the controller dir


stuggi (approved):
/lgtm

PR #115, opened 2024-05-08
Check for valid service override endpoint type on create and update

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/505


d34dh0r53 (approved):
LGTM

d34dh0r53 (changes requested):
It looks like the `ValidateRoutedOverrides` function is missing.

xek (approved):
LGTM

PR #145, opened 2024-07-29
Update getSecret to use VerifySecret

mauricioharley:


controllers/barbicanapi_controller.go
```diff
@@ -197,15 +198,19 @@ func (r *BarbicanAPIReconciler) getSecret(
 	secretName string,
 	envVars *map[string]env.Setter,
 ) (ctrl.Result, error) {
-	hash, result, err := secret.VerifySecret(ctx, types.NamespacedName{Name: instance.Spec.Secret, Namespace: instance.Namespace}, []string{"AdminPassword"}, h.GetClient(), time.Second*10)
+	expectedField := "transport_url"
+	if strings.HasSuffix(secretName, "osp-secret") {
```
dmendiza, line 202: We don't want to hard-code `"osp-secret"` here, because that value is configurable in the API spec.  This should be updated to use the value provided in the spec `instance.Spec.Secret`, which already defaults to `"osp-secret"`:

https://github.com/openstack-k8s-operators/barbican-operator/blob/main/api/v1beta1/common_types.go#L40

controllers/barbicanapi_controller.go
```diff
@@ -197,15 +198,19 @@ func (r *BarbicanAPIReconciler) getSecret(
...
+		expectedField = "AdminPassword"
```
dmendiza, line 203: Although you are just reusing this existing string, hard-coding  `"AdminPassword"` here is likely a bug.  Every secret used, and they keys needed within that secret to access the data we need should be defined in the spec, not hard-coded into the controller like it is currently.

This value should probably be replaced with `instance.Spec.PasswordSelectors.Service`, which is the password that we need for the Barbican service user.  The default value for that is `"BarbicanPassword"`, not `"AdminPassword"`.

https://github.com/openstack-k8s-operators/barbican-operator/blob/main/api/v1beta1/common_types.go#L107-L109

It seems things are currently working because install_yamls reuses the same password for both of these fields in `"osp-secret"`, but these should be different in a real world deployment.

https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-input-kustomize.sh#L65-L67

Given that we care about different keys depending on which secret is being retrieved, then I think it makes sense that the list of keys we care about should be passed into this method alongside the name of the relevant secret.

dmendiza (changes requested):

controllers/barbicanapi_controller.go
stuggi: wouldn't it be easier to maintain to have another new parameter to the func where the caller passes in the `[]string{}` to check in `VerifySecret()`. So for the `instance.Spec.Secret` it would be`[]string{instance.Spec.PasswordSelectors.Service}` and for the transportURL `[]string{"transport_url"}` . I think this would make it easier to understand and won't require this condition

dmendiza: @stuggi that's the same suggestion I made in the other comment below. :point_down:

stuggi: oh right, sorry haven't read it close enough

PR #153, opened 2024-08-19
Adding nodeSelector functionality

mauricioharley:
The previous Barbican operator's code was not taking actions when the `nodeSelector` attribute was specified in the control plane. This patch addresses the missing part.

PR #159, opened 2024-09-09
[sample] comment optional nodeSelector

stuggi:
When other services deploy barbican as a dependency in CI using the barbican sample the nodeselector functionality, which was added recently with
https://github.com/openstack-k8s-operators/barbican-operator/pull/153 prevents the pods from being able to be scheduled. Therefore this change will comment these optional settings.


abays (approved):
/lgtm

PR #162, opened 2024-09-20
Adding initial support for Eviden Trustway HSM

mauricioharley:
The shell script automates the necessary steps to include Eviden's Trustway HSM (Hardware Security Module) client software into the container images for Barbican API and Barbican Worker.

The code is the initial support spec to include the same vendor and product.


hack/build_custom_image-eviden.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+# This script creates a custom container image for both, Barbican API and Barbican Worker,
+# including the HSM vendor's client software.
+# Vendor:  Eviden
+
+if [ "$#" -ne 5 ]; then
+    echo "Usage: $0 <registry_host> <namespace> <barbican-api_image_tag> <barbican-worker_image_tag> <eviden_iso_file>"
+    exit 1
+fi
```
vakwetu, line 9: Shouldn't we have two registry hosts here - a source registry host (which should default to whatever the official RH registry is)  and a destination registry host - which would be your private registry?  These won't necessarily be the same.

vakwetu (changes requested):

hack/build_custom_image-eviden.sh
mauricioharley: Understood.  I was looking for clarifying this but couldn't find any evidence.  Will fix it in the next push.

api/bases/barbican.openstack.org_barbicanapis.yaml
```diff
@@ -35,6 +35,10 @@ spec:
           spec:
             description: BarbicanAPISpec defines the desired state of BarbicanAPI
             properties:
+              clientCertificate:
```
dmendiza, line 38: Maybe we should consider grouping these into their own key.  e.g. 

    spec:
      ...
      properties:
        pkcs11:
          clientCertificate: $CERT_CONTENTS_HERE

mauricioharley: Got it.  $CERT_CONTENTS_HERE will receive the OpenShift secret name that actually stores the certificate.

This file was automatically generated by a pre-commit hook.  I'll check how this could be translated into the specification file.

vakwetu (changes requested):
There are a whole bunch of changes here which are for the CRD, for which we decided to do differently in the lunasa patch.  I had thought that this PR had been modified to only contain the scripts.  But I still see them here.

vakwetu:
This PR seems to be stale and is likely no longer relevant.  I recall we chose not to add many of these parameters into the CRDs.

Not sure why this popped up again - but likely this should be closed?

abays:
> This PR seems to be stale and is likely no longer relevant. I recall we chose not to add many of these parameters into the CRDs.
> 
> Not sure why this popped up again - but likely this should be closed?

We just updated some config in Prow this morning that triggered a bunch of old PRs.

mauricioharley:
Abandoned in detriment of another approach.

PR #168, opened 2024-10-09
Barbican Support for Luna HSM

mauricioharley:
This patch is a Work in Progress to support Thales Luna HSMs on Barbican v18.

It templates the Luna HSM configuration into Barbican variables.


api/v1beta1/common_types.go
```diff
@@ -101,6 +101,83 @@ type BarbicanComponentTemplate struct {
 	NetworkAttachments []string `json:"networkAttachments,omitempty"`
 }
 
+// BarbicanLunaTemplate - Variables specific to Thales' Luna HSMs
```
vakwetu, line 104: Lets avoid using company names like "Thales".  The companies have changed a bunch!

api/v1beta1/common_types.go
```diff
@@ -101,6 +101,83 @@ type BarbicanComponentTemplate struct {
...
+	LunaTokenSerialNumber string `json:"lunaTokenSerialNumber"`
+
+        // +kubebuilder:validation:Optional
+	// Token label used to identify the token to be used.  Required when
+	// token_serial_number is not specified.
+	LunaTokenLabel string `json:"lunaTokenLabel"`
+
+        // +kubebuilder:validation:Optional
+	// The OpenShift secret containing the password to login to PKCS11 session
+	LunaPassword string `json:"lunaPassword"`
```
vakwetu, line 140: LunaPassword shouldn't be stored here.  We're going to put it in openstack-secret.  Instead you should be adding to the PasswordSelector.

api/v1beta1/common_types.go
```diff
@@ -101,6 +101,83 @@ type BarbicanComponentTemplate struct {
 	NetworkAttachments []string `json:"networkAttachments,omitempty"`
 }
 
+// BarbicanLunaTemplate - Variables specific to Thales' Luna HSMs
+type BarbicanLunaTemplate struct {
```
vakwetu, line 105: This structure should be added the relevant specs - ie. the BarbicanApiSpec and the BarbicanWorkerSpec

vakwetu (commented):
Next step of course is to take what you have here and add to the relevant specs.

Then you need to find where the config  files are generated/templated and make the changes.  See https://github.com/openstack-k8s-operators/keystone-operator/pull/479/files# as an example of what Dave is doing on the federation side.

api/v1beta1/common_types.go
```diff
@@ -101,6 +101,83 @@ type BarbicanComponentTemplate struct {
...
+
+        // +kubebuilder:validation:Optional
+        // The HSM certificates. The map's key is the HSM's IP address and
+        // the value is the OpenShift secret storing the certificate.
+        LunaHSMCertificates map[string]string `json:"lunaHsmCertificates"`
+
+        // +kubebuilder:validation:Optional
+        // The OpenShift secret storing the client certificate plus its key
+        LunaClientCertificate string `json:"lunaClientCertificate"`
+
```
vakwetu, line 121: If I recall correctly, we've decided that we're going to take these files and put them into a secret which is to be mounted.  So that means that what you really need here is a parameter that contains the name of the secret to be mounted.

vakwetu: Actually, now that I think about it, you're going to need the secret - but also you're going to need the path names for the certs.

api/v1beta1/common_types.go
mauricioharley: In another PR, there was a decision towards storing server and client certificates and key in a separate secret, not in `osp-secret`. What is the final decision then?

api/v1beta1/common_types.go
mauricioharley: I'll create another struct called `Pkcs11Template` that will encompass this one. It will then be added to `BarbicanApiSpec` and `BarbicanWorkerSpec`.

Agreed, or do you have another suggestion?

api/v1beta1/common_types.go
mauricioharley: Kindly observe the comment right above the variable name. It's not meant to directly provide the certificate, but actually the secret that contains the certificate.

api/v1beta1/barbicanapi_types.go
```diff
@@ -64,6 +64,17 @@ type BarbicanAPISpec struct {
 
 	BarbicanAPITemplate `json:",inline"`
 
+	PKCS11 BarbicanPKCS11Template `json:"pkcs11,omitempty"`
```
vakwetu, line 67: Need some kubebuilder entries above this.   something that says optional.

api/v1beta1/barbicanapi_types.go
```diff
@@ -64,6 +64,17 @@ type BarbicanAPISpec struct {
...
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:default=Enum=simple_crypto
+	EnabledSecretStores []string `json:"enabledSecretStores"`
```
vakwetu, line 72: Is this right?   default=["simple_crypto"]   ?

api/v1beta1/barbicanworker_types.go
```diff
@@ -44,7 +44,19 @@ type BarbicanWorkerSpec struct {
 	BarbicanTemplate `json:",inline"`
 
 	BarbicanWorkerTemplate `json:",inline"`
-	DatabaseHostname       string `json:"databaseHostname"`
+
+	PKCS11 BarbicanPKCS11Template `json:"pkcs11,omitempty"`
```
vakwetu, line 48: kubebuilder options needed

controllers/barbicanapi_controller.go
```diff
@@ -326,6 +327,45 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
 		"EnableSecureRBAC": instance.Spec.EnableSecureRBAC,
 	}
 
+	// Checking if there's an HSM.
+	pkcs11 := instance.Spec.PKCS11
```
vakwetu, line 331: this will fail when there is no PKCS11 element because this is optional.  You need to check  
if (len(instance.Spec.PKCS11) > 0)  or check for !=nil

As a result , you can use this ^^ and you dont need the HSMEnabled flag.

controllers/barbicanapi_controller.go
```diff
@@ -326,6 +327,45 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
+	if pkcs11.HSMEnabled {
+		hsmLoginSecret, _, err := secret.GetSecret(ctx, h, pkcs11.HSMLogin, instance.Namespace)
+		if err != nil {
+			return err
+		}
+		hsmCertificatesSecret, _, err := secret.GetSecret(ctx, h, maps.Keys(pkcs11.HSMCertificates)[0], instance.Namespace)
```
vakwetu, line 337: You do not need this secret here.  This is mounted by the volumes/ volume mounts

controllers/barbicanapi_controller.go
```diff
@@ -326,6 +327,45 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
+		if err != nil {
+			return err
+		}
+		templateParameters["HSMEnabled"] = pkcs11.HSMEnabled
```
vakwetu, line 341: This should be true

controllers/barbicanapi_controller.go
```diff
@@ -326,6 +327,45 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
+				templateParameters["HSMServerCertfile"] = certfile
+				templateParameters["HSMServerCertificate"] = certificate
+			} else if strings.HasSuffix(certfile, "Key.pem") {
+				templateParameters["HSMClientKeyFile"] = certfile
+				templateParameters["HSMClientKey"] = certificate
+			} else if !(strings.HasSuffix(certfile, "File.pem")) { // Excluding "CAFile.pem".
+				templateParameters["HSMClientCertfile"] = certfile
+				templateParameters["HSMClientCertificate"] = certificate
+			}
+		}
```
vakwetu, line 366: remove this

controllers/barbicanworker_controller.go
```diff
@@ -278,6 +280,45 @@ func (r *BarbicanWorkerReconciler) generateServiceConfigs(
 		"SimpleCryptoKEK": string(simpleCryptoSecret.Data[instance.Spec.PasswordSelectors.SimpleCryptoKEK]),
 	}
 
+	// Checking if there's an HSM.
```
vakwetu, line 283: see some notes as above

templates/barbican/config/00-default.conf
```diff
@@ -44,15 +44,38 @@ enable = true
 
 [secretstore]
 enable_multiple_secret_stores = true
-stores_lookup_suffix = simple_crypto
+stores_lookup_suffix = {{ .enabledSecretStores }}
```
vakwetu, line 47: You need to add logic to pass this parameter in -- string.join(instance.Spec.EnabledSecretStores, ",")

vakwetu: Same for the *CryptoEnabled .. etc. as in our google doc.

api/v1beta1/barbicanapi_types.go
```diff
@@ -64,6 +64,17 @@ type BarbicanAPISpec struct {
...
+
+	// +kubebuilder:validation:Optional
```
vakwetu, line 74: This is required - and if we can get an enum of values, that would be awesome.  It is a single string that must be one of "simple_crypto", "pkcs11" ..

templates/barbican/config/barbican-api-config.json
```diff
@@ -77,6 +77,14 @@
       "perm": "0600",
       "optional": true,
       "merge": true
+    },
+    {
+      "source": "/var/lib/config-data/default/Crystoki.conf",
+      "dest": "{{ .HSMLibraryPath }}/config/Crystoki.conf",
+      "owner": "barbican",
+      "perm": "0600",
+      "optional": true,
```
vakwetu, line 86: This will work for now until we learn how to do this properly, but the way things are right now - this will always be added.  why?  because Crystoki.conf is in templates/barbican/config/ and therefore will always be templated - even if we're not using  a luna

api/v1beta1/common_types.go
```diff
@@ -101,6 +101,79 @@ type BarbicanComponentTemplate struct {
...
+	HSMIPAddress string `json:"hsmIpAddress"`
+
+	// +kubebuilder:validation:Optional
+	// The IP address of the client connecting to the HSM (X.Y.Z.K)
+	HSMClientAddress string `json:"hsmClientAddress"`
+
+        // +kubebuilder:validation:Optional
+        // The HSM certificates. The map's key is the OpenShift secret storing the certificate, and
+	// the value is the mounting point (e.g., "luna-certificates": "/usr/local/luna/config/certs").
+        HSMCertificates map[string]string `json:"hsmCertificates"`
```
stuggi, line 169: can there be multiple hsm certificates involved? because if I read the PR correct, only one would be mounted .  If if only one cert secret is used, can't we just mount them to a fixed patch, instead of making the path configurable?
If we need it to be configurable, why not have a list of a sub struct type with `secretName` and `mountPath` ?

stuggi: also what is the expected content of the secret? specific keys for referencing cert/key?

api/v1beta1/barbican_types.go
```diff
@@ -22,10 +22,16 @@ import (
...
 )
 
+// Supported HSM models
+var HSMTypes = []string{"luna", "ncipher", "trustway"}
+
 const (
 	// DbSyncHash hash
 	DbSyncHash = "dbsync"
 
+	// DbSyncHash hash
```
xek, line 32: P11PrepHash

config/samples/pods_with_pkcs11-luna.yaml
```diff
@@ -0,0 +1,85 @@
...
+    hsmLoginSecret: my_luna_login_secret
+    hsmCertificatesSecret: my_luna_certs_secret
+    hsmCertificatesMountPoint: /usr/local/luna/config/certs
+    hsmKeyWrapMechanism: CKM_AES_CBC_PAD
+  barbicanAPI:
+    globalDefaultSecretStore: pkcs11
+    enabledSecretStores:
+      - simple_crypto
+      - pkcs11
+    pkcs11:
```
xek, line 47: We should be able to define this in one place

controllers/barbican_common.go
```diff
@@ -52,12 +57,94 @@ func GenerateConfigsGeneric(
...
+	namespace string,
+) (map[string]interface{}, error) {
+	tempMap := map[string]interface{}{}
+	hsmLoginSecret, _, err := oko_secret.GetSecret(ctx, h, pkcs11.HSMLoginSecret, namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	// Checking if at least one of the following parameters has been provided.
+	if len(pkcs11.HSMTokenSerialNumber) == 0 && len(pkcs11.HSMTokenLabels) == 0 && len(pkcs11.HSMSlotId) == 0 {
```
xek, line 111: Investigate whether we could use webhooks for validation

controllers/barbicanworker_controller.go
```diff
@@ -278,6 +280,38 @@ func (r *BarbicanWorkerReconciler) generateServiceConfigs(
 		"SimpleCryptoKEK": string(simpleCryptoSecret.Data[instance.Spec.PasswordSelectors.SimpleCryptoKEK]),
 	}
 
+	for key, value := range templateParameters {
```
vakwetu, line 283: Remove this logging stuff

controllers/barbicanworker_controller.go
```diff
@@ -278,6 +280,38 @@ func (r *BarbicanWorkerReconciler) generateServiceConfigs(
...
+	// Set secret store parameters
+	secretStoreTemplateMap, err := GenerateSecretStoreTemplateMap(
+		instance.Spec.EnabledSecretStores,
+		instance.Spec.GlobalDefaultSecretStore)
+	if err != nil {
+		return err
+	}
+	maps.Copy(templateParameters, secretStoreTemplateMap)
+
+	for key, value := range templateParameters {
```
vakwetu, line 296: Remove this logging stuff

controllers/barbicanworker_controller.go
```diff
@@ -278,6 +280,38 @@ func (r *BarbicanWorkerReconciler) generateServiceConfigs(
...
+		pkcs11TemplateMap, err := GeneratePKCS11TemplateMap(
+			ctx, h, instance.Spec.PKCS11, instance.Namespace)
+		if err != nil {
+			return err
+		}
+		maps.Copy(templateParameters, pkcs11TemplateMap)
+		instance.Spec.PKCS11.HSMEnabled = true
+	}
+
+	for key, value := range templateParameters {
```
vakwetu, line 311: Remove this logging stuff

templates/barbican/config/00-default.conf
```diff
@@ -33,26 +33,71 @@ topic = barbican_notifications
...
 
 [simple_crypto_plugin]
 plugin_name = Software Only Crypto
-{{ if (index . "SimpleCryptoKEK") }}
+{{- if (index . "SimpleCryptoKEK") }}
 kek = {{ .SimpleCryptoKEK }}
+{{- end }}
 {{ end }}
+
+{{- if and (index . "HSMEnabled") .HSMEnabled }}
```
vakwetu, line 68: change to pk11Enabled

templates/barbican/config/Chrystoki.conf
```diff
@@ -0,0 +1,39 @@
+{{ if and (index . "HSMEnabled") .HSMEnabled }}
```
vakwetu, line 1: if hsm enabled and type is luna

templates/barbican/config/barbican-api-config.json
```diff
@@ -77,6 +77,14 @@
       "perm": "0600",
       "optional": true,
       "merge": true
+    },
+    {
+      "source": "/var/lib/config-data/default/Chrystoki.conf",
+      "dest": "//usr/local/luna/Chrystoki.conf",
```
vakwetu, line 83: remove double slash

templates/barbican/config/barbican-worker-config.json
```diff
@@ -27,6 +27,14 @@
         "owner": "barbican",
         "perm": "0755",
         "optional": true
+      },
+      {
+	"source": "/var/lib/config-data/default/Chrystoki.conf",
+	"dest": "//usr/local/luna/Chrystoki.conf",
```
vakwetu, line 33: double slash!

api/v1beta1/common_types.go
```diff
@@ -95,6 +107,130 @@ type BarbicanComponentTemplate struct {
...
 
+// +kubebuilder:validation:Enum=simple_crypto;pkcs11
+// This SecretStore type is used by the EnabledSecretStores variable inside the specification.
+type SecretStore string
+
+// BarbicanPKCS11Template - Includes all common HSM properties
+type BarbicanPKCS11Template struct {
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:default=false
+	HSMEnabled bool `json:"hsmEnabled"`
```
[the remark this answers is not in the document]
vakwetu: removed

config/samples/pods_with_pkcs11-luna.yaml
vakwetu: yes - this works!  removing the extras

api/v1beta1/common_types.go
vakwetu: This has been modified.  We can have multiple certs but they will all be contained  in a single secret - which will be mounted to a specific directory (mountpoint).  The contents of the secret will be all the files that will be created in that directory.

controllers/barbican_common.go
vakwetu: ack - lets leave this to a follow-on patch.  I started to look at this, but its a little more involved than I initially thought.

vakwetu: done

vakwetu:
Looks like kuttl tests are passing.  but the deploy-test there is failing while waiting on galera.

controllers/barbican_controller.go
```diff
@@ -919,11 +949,53 @@ func (r *BarbicanReconciler) reconcileInit(
 	}
 	instance.Status.Conditions.MarkTrue(condition.DBSyncReadyCondition, condition.DBSyncReadyMessage)
 
+	//
+	// run Barbican p11-prep if needed
+	//
+	if slices.Contains(instance.Spec.EnabledSecretStores, "pkcs11") {
```
xek, line 955: We should have an else here, update the status condition to true with a message like:

"P11 Prep job not run"

This is why it's failing to create a deployment after kuttl tests run, since by not setting it, "Ready" condition is also not set in [1], and consequently the openstack control plane is also not ready.

[1] https://github.com/openstack-k8s-operators/barbican-operator/pull/168/files#diff-cb25cf601c065f6057f4ae086a63e5796d9790074eb5b842f2b9169fed00cd58L447

xek (changes requested):

hack/crd-schema-checker.sh
```diff
@@ -16,6 +16,7 @@ for crd in config/crd/bases/*.yaml; do
     mkdir -p "$(dirname "$TMP_DIR/$crd")"
     git show "$BASE_REF:$crd" > "$TMP_DIR/$crd"
     $CHECKER check-manifests \
+        --disabled-validators=NoBools,NoNewRequiredFields \
```
vakwetu, line 19: OK - so this was just for testing.  We want to remove this and then explicitly waive the test failures.

xek (approved):
/lgtm

vakwetu:
Just to add a comment here about the CRD checker.

This patch was accidentally merged without removing the changes to the crd-schema-checker.  We will re-enable them in a follow-on commit.

The checker flagged a bunch of bools that were added in the pkcs11 spec.  These bools are in fact really bools that would be set as such in the barbican.conf file.  So the errors here would have been waived.

Further, the checker flagged a bunch of new required fields in the PKCS11 struct.  This is OK because the PKCS11 struct itself is new and it is included into our existing structs as an optional pointer to a PKCS11 struct.  Existing CRs will not have this field defined, and upon update, will instantiate that pointer as a nil value.  Which won't trigger any issues with required fields in the struct.

PR #182, opened 2024-11-11
Fix CustomServiceConfig barbican configuration

Deydra71:
This PR adds support for the `CustomServiceConfig` field in the Barbican CRD, allowing custom configurations for the Barbican API, worker, and keystone-listener services.

Example:
```yaml
barbican:
  template:
    databaseInstance: openstack
    secret: osp-secret
    barbicanAPI:
      customServiceConfig: |
        [DEFAULT]
        max_allowed_secret_in_bytes = 1048576
      replicas: 1
 ```

The custom config file is now mounted as a  `02-service-custom.conf` file to `/var/lib/config-data/default/`  location and copied to `/etc/barbican/barbican.conf.d/` as a `custom.conf` file.


xek (approved):
Thanks for fixing this!

PR #190, opened 2024-11-28
Functional tests for HSM Support

mauricioharley:
Functional tests for the Luna HSM model support.

### DIFF
--- a/tests/functional/barbican_controller_test.go
+++ b/tests/functional/barbican_controller_test.go
@@ -12,7 +12,7 @@ import (
 	. "github.com/openstack-k8s-operators/lib-common/modules/common/test/helpers"
 
 	barbicanv1beta1 "github.com/openstack-k8s-operators/barbican-operator/api/v1beta1"
-	"github.com/openstack-k8s-operators/barbican-operator/controllers"
+	controllers "github.com/openstack-k8s-operators/barbican-operator/controllers"
 	"github.com/openstack-k8s-operators/barbican-operator/pkg/barbican"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	mariadb_test "github.com/openstack-k8s-operators/mariadb-operator/api/test/helpers"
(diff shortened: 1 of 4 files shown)


hack/crd-schema-checker.sh
```diff
@@ -16,6 +16,7 @@ for crd in config/crd/bases/*.yaml; do
     mkdir -p "$(dirname "$TMP_DIR/$crd")"
     git show "$BASE_REF:$crd" > "$TMP_DIR/$crd"
     $CHECKER check-manifests \
+        --disabled-validators=NoBools,NoNewRequiredFields \
```
gibizer, line 19: I strongly disagree to disable NoNewRequiredFields as we learned in the hard way that adding new required field will block update on the customer side.

gibizer (dismissed):

api/v1beta1/common_types.go
```diff
@@ -49,6 +49,19 @@ type BarbicanTemplate struct {
 	// +kubebuilder:validation:Required
 	// ServiceAccount - service account name used internally to provide Barbican services the default SA name
 	ServiceAccount string `json:"serviceAccount"`
+
+        // +kubebuilder:validation:Optional
+        PKCS11 *BarbicanPKCS11Template `json:"pkcs11,omitempty"`
```
gibizer, line 54: So this is a new optional struct field. 
If the user is not passing anything here then the field will be defaulted to nil and therefore the required struct fields does not cause a problem as the struct is not instantiated. 
If the user is passing a value to this field then the user will need to provide at least all the required struct fields as well. 

This feels like an exception over the rule of not adding new required fields, as the struct field wrapping the new required field is i) introduced by the same change so there is no CRs on the field with a value for this field ii) the struct field is introduced as an optional, *struct field, defaulted to nil.

I suggest to keep the crd check in place but ignore this false positive when merging the PR (by overriding the CI result) if the rest of the pre-commit checks are passing.

hack/crd-schema-checker.sh
mauricioharley: Just saw this comment.  Will address it on the next push.

tests/functional/barbican_controller_test.go
```diff
@@ -422,6 +423,150 @@ var _ = Describe("Barbican controller", func() {
...
+				corev1.ConditionTrue,
+			)
+
+			BarbicanAPIExists(barbicanTest.Instance)
+
+			d := th.GetDeployment(barbicanTest.BarbicanAPI)
+			// Check the resulting deployment fields
+			Expect(int(*d.Spec.Replicas)).To(Equal(1))
+
+			Expect(d.Spec.Template.Spec.Volumes).To(HaveLen(4))
```
vakwetu, line 469: I'm ok with these coming in as subsequent patches, but I'm adding the comments here for context.
It would be good to confirm that one of these volumes is the lunaVolume as populated for by the lunaCertSecret and mounted to the right place. You do this below for the p1Prep job

tests/functional/barbican_controller_test.go
```diff
@@ -422,6 +423,150 @@ var _ = Describe("Barbican controller", func() {
...
+				}
+			}
+
+			volume := th.GetJob(barbicanTest.BarbicanP11Prep).Spec.Template.Spec.Containers[0].VolumeMounts[elemNum].Name
+			mountPath := th.GetJob(barbicanTest.BarbicanP11Prep).Spec.Template.Spec.Containers[0].VolumeMounts[elemNum].MountPath
+
+			Eventually(func(g Gomega) {
+				g.Expect(volume).To(Equal(barbican.LunaVolume))
+				g.Expect(mountPath).To(Equal(HSMCertificatesMountPoint))
+			}, timeout, interval).Should(Succeed())
```
vakwetu, line 566: There is also a scripts volume that is new here and should probably be checked.

tests/functional/barbican_controller_test.go
```diff
@@ -422,6 +423,150 @@ var _ = Describe("Barbican controller", func() {
...
+				ContainSubstring("dummy-data"))
+			confClient := certsSecret.Data[HSMClientAddress+"Client.pem"]
+			Expect(confClient).To(
+				ContainSubstring("dummy-data"))
+			confKey := certsSecret.Data[HSMClientAddress+"Client.key"]
+			Expect(confKey).To(
+				ContainSubstring("dummy-data"))
+		})
+
+		It("Verifies if Chrystoki.conf and 00-default.conf have the right contents.", func() {
```
vakwetu, line 519: You should probably repeat this test for the BarbicanAPIConfigSecret

vakwetu (commented):
A good start generally.  Just a few suggestions of things to add.

vakwetu:
recheck

vakwetu (commented):
All the changes you have made look pretty good.  There are a bunch of unrelated changes though.  Please rebase first.

vakwetu (approved):
LGTM

PR #199, opened 2025-01-13
Make the pkcs11 config more generic

vakwetu:
Made the pkcs11 more generic so that hopefully we will be able to support many HSMs and configurations without
too many changes.

Jira: https://issues.redhat.com/browse/OSPRH-12972


pkg/barbican/p11_prep.go
```diff
@@ -40,7 +40,7 @@ func P11PrepJob(instance *barbicanv1beta1.Barbican, labels map[string]string, an
 
 	// add any HSM volumes
 	p11Volumes = append(p11Volumes, GetHSMVolumes(*instance.Spec.PKCS11)...)
-	p11Mounts = append(p11Mounts, GetHSMVolumeMounts(*instance.Spec.PKCS11)...)
+	p11Mounts = append(p11Mounts, GetHSMVolumeMounts()...)
```
stuggi, line 43: these volumes/mounts will fail if `*instance.Spec.PKCS11` is nil or the data secret does note exist. see my comment bellow for adding an input validation.

api/v1beta1/barbican_webhook.go
```diff
@@ -141,13 +141,6 @@ func (r *BarbicanSpec) ValidateCreate(basePath *field.Path) field.ErrorList {
 			allErrs = append(allErrs, field.Required(basePath.Child("PKCS11"),
```
stuggi, line 141: we should have the same check on `ValidateUpdate` if someone enables P11 as a day2.

controllers/barbican_controller.go
```diff
@@ -670,12 +670,13 @@ func (r *BarbicanReconciler) generateServiceConfig(
 
 	// Set pkcs11 parameters
 	if slices.Contains(instance.Spec.EnabledSecretStores, "pkcs11") {
```
stuggi, line 672: Adding it here since you can not comment at non changed lines. 

* we should add an input check for `ClientDataSecret` when specified, like we do with the password secret [1] and add the hash so that when the content changes the deployments get restarted
* index the secret field, like with the cert secrets [2][3], to reconcile when the content of the secret changes. we'd have to add this to all controllers which should reconcile when the content changes.

[1] https://github.com/openstack-k8s-operators/barbican-operator/blob/main/controllers/barbican_controller.go#L258-L278
[2] https://github.com/openstack-k8s-operators/barbican-operator/blob/main/controllers/barbican_controller.go#L561
[3] https://github.com/openstack-k8s-operators/barbican-operator/blob/main/controllers/barbicanapi_controller.go#L899-L909

api/v1beta1/common_types.go
```diff
@@ -112,120 +112,20 @@ type BarbicanComponentTemplate struct {
 // This SecretStore type is used by the EnabledSecretStores variable inside the specification.
 type SecretStore string
```
stuggi, line 113: lets add the following const here and we could replace the hard coded strings in the controllers with those, like `if slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStorePKCS11) {`

```
const (
	// SecretStoreSimpleCrypto -
	SecretStoreSimpleCrypto SecretStore = "simple_crypto"

	// SecretStorePKCS11 -
	SecretStorePKCS11 SecretStore = "pkcs11"
)
```

pkg/barbicanapi/deployment.go
```diff
@@ -107,7 +107,7 @@ func Deployment(
 	// Add PKCS11 volumes
 	if slices.Contains(instance.Spec.EnabledSecretStores, "pkcs11") {
```
stuggi, line 108: like mentioned above could change `pkcs11` => `barbicanv1beta1.SecretStorePKCS11` when introduce the consts

pkg/barbicanworker/deployment.go
```diff
@@ -83,7 +83,7 @@ func Deployment(
 	// Add PKCS11 volumes
 	if slices.Contains(instance.Spec.EnabledSecretStores, "pkcs11") {
```
stuggi, line 84: like mentioned above could change pkcs11 => barbicanv1beta1.SecretStorePKCS11 when introduce the consts

templates/barbican/config/barbican-api-config.json
```diff
@@ -78,14 +78,16 @@
       "optional": true,
       "merge": true
     },
```
stuggi, line 80: this will not be valid json when `P11ClientDataPath` is not set and this element ends with `,`. could just move the conditional entry somewhere above to not be the last one.

templates/barbican/config/barbican-api-config.json
```diff
@@ -78,14 +78,16 @@
...
+    {{- if (index . "P11ClientDataPath") }}
```
stuggi, line 81: if we default `P11ClientDataPath` to some value if `ClientDataPath` is not set, then we could just add that entry with out the if, we already set `"optional": true,` so its fine to not have files

templates/barbican/config/barbican-api-config.json
```diff
@@ -78,14 +78,16 @@
...
     {
-      "source": "/var/lib/config-data/default/Chrystoki.conf",
-      "dest": "/usr/local/luna/Chrystoki.conf",
+      "source": "/etc/p11-client/*",
```
stuggi, line 83: should be `/var/lib/config-data/hsm`, right ?

templates/barbican/config/barbican-p11-prep-config.json
```diff
@@ -13,14 +13,16 @@
 	"owner": "barbican",
 	"perm": "0600"
       },
+      {{- if (index . "P11ClientDataPath") }}
       {
-        "source": "/var/lib/config-data/default/Chrystoki.conf",
-        "dest": "/usr/local/luna/Chrystoki.conf",
+        "source": "/etc/p11-client/*",
```
stuggi, line 18: should be /var/lib/config-data/hsm, right ?

templates/barbican/config/barbican-p11-prep-config.json
```diff
@@ -13,14 +13,16 @@
 	"owner": "barbican",
 	"perm": "0600"
       },
+      {{- if (index . "P11ClientDataPath") }}
```
stuggi, line 16: same as above

api/v1beta1/common_types.go
```diff
@@ -112,120 +112,20 @@ type BarbicanComponentTemplate struct {
...
+        // These will be mounted to /etc/hsm-client
+        ClientDataSecret string `json:"clientDataSecret"`
 
-        // +kubebuilder:validation:Optional
-        // +kubebuilder:default=false
-        // Set os_locking_ok
-	OSLockingOK  bool `json:"OSLockingOK"`
+        // +kubebuilder:validation:Required
+        // Location to which kolla will copy the data in ClientDataSecret.
+        ClientDataPath string `json:"clientDataPath"`
```
stuggi, line 128: wondering if we could make it optional and default it to e.g. `/etc/hsm-client`? see my comment on the barbican kolla file. would remove the req to have the condition in it.

controllers/barbican_controller.go
stuggi: same applies to the `LoginSecret` and that the entry for the `P11CryptoLogin` selector exist. should also add the hash,

api/v1beta1/barbican_webhook.go
vakwetu: Done

api/v1beta1/common_types.go
vakwetu: Dine

api/v1beta1/common_types.go
vakwetu: Done

pkg/barbican/p11_prep.go
vakwetu: Done

templates/barbican/config/barbican-api-config.json
vakwetu: No longer conditional

controllers/barbican_controller.go
```diff
@@ -555,22 +538,32 @@ func (r *BarbicanReconciler) reconcileDelete(ctx context.Context, instance *barb
...
-	passwordSecretField     = ".spec.secret"
-	caBundleSecretNameField = ".spec.tls.caBundleSecretName"
-	tlsAPIInternalField     = ".spec.tls.api.internal.secretName"
-	tlsAPIPublicField       = ".spec.tls.api.public.secretName"
+	passwordSecretField      = ".spec.secret"
+	caBundleSecretNameField  = ".spec.tls.caBundleSecretName"
+	tlsAPIInternalField      = ".spec.tls.api.internal.secretName"
+	tlsAPIPublicField        = ".spec.tls.api.public.secretName"
+	p11LoginSecretField      = ".spec.pkcs11.loginSecret"
+	p11ClientDataSecretField = ".spec.pkcs11.clientDataSecret"
```
stuggi, line 546: there is a bit more todo to index the field, check here for reference https://github.com/openstack-k8s-operators/barbican-operator/blob/main/controllers/barbicanapi_controller.go#L863-L873 . we'd need the same for the new fields in all controllers.
now that we have the job in the barbican controller, we'd also have to add the equivalent to the barbican controller, or move the job to the api controller of that makes sense.

controllers/barbicanworker_controller.go
```diff
@@ -291,13 +301,14 @@ func (r *BarbicanWorkerReconciler) generateServiceConfigs(
 	maps.Copy(templateParameters, secretStoreTemplateMap)
 
 	// Set pkcs11 parameters
-	if slices.Contains(instance.Spec.EnabledSecretStores, "pkcs11") {
-		pkcs11TemplateMap, err := GeneratePKCS11TemplateMap(
-			ctx, h, *instance.Spec.PKCS11, instance.Namespace)
+	if slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStorePKCS11) {
```
stuggi, line 304: `&& instance.Spec.PKCS11 != nil`

controllers/barbican_controller.go
```diff
@@ -252,52 +252,35 @@ func (r *BarbicanReconciler) reconcileNormal(ctx context.Context, instance *barb
...
-	if err != nil {
-		if k8s_errors.IsNotFound(err) {
-			Log.Info(fmt.Sprintf("simple Crypto backend secret %s not found", instance.Spec.SimpleCryptoBackendSecret))
-			instance.Status.Conditions.Set(condition.FalseCondition(
-				condition.InputReadyCondition,
-				condition.RequestedReason,
-				condition.SeverityInfo,
-				condition.InputReadyWaitingMessage))
-			return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
+	if len(instance.Spec.EnabledSecretStores) == 0 || slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStoreSimpleCrypto) {
```
stuggi, line 262: should this be `> 0` instead of `==`?

stuggi, line 262: I think with `len(instance.Spec.EnabledSecretStores) == 0` would make the `SimpleCryptoBackendSecret` a required secret, which would not be there on existing deployments and reconcile will probably stop here as the secret is missing. is that what we want? just thinking about how that'd impact the update of an existing env

stuggi: was the same with the commit which introduced the check of SimpleCryptoBackendSecret, but something we need to consider when update from fr1

stuggi: maybe should be `(len(instance.Spec.EnabledSecretStores) == 0 && SimpleCryptoBackendSecret != "") || ...`?

stuggi: ah no that won't work as it defaults to the `osp-secret`.

vakwetu: SimpleCryptoBackend was the only thing available till now.  Any existing deployments would have to have had a SimpleCryptoSecret defined (or defaulted to osp-secret).  So I don't think we need to worry about existing deployments.

templates/barbican/config/barbican-api-config.json
vakwetu: Done

templates/barbican/config/barbican-api-config.json
vakwetu: Done

templates/barbican/config/barbican-p11-prep-config.json
vakwetu: Done

templates/barbican/config/barbican-p11-prep-config.json
vakwetu: Done

controllers/barbican_controller.go
```diff
@@ -252,52 +252,33 @@ func (r *BarbicanReconciler) reconcileNormal(ctx context.Context, instance *barb
...
+	// check for PKCS11 secret holding the PKCS11 Crypto Login
+	if slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStorePKCS11) {
+		ctrlResult, err = r.verifySecret(ctx, helper, instance, instance.Spec.PKCS11.LoginSecret, []string{instance.Spec.PasswordSelectors.P11CryptoLogin}, &configVars)
+		if err != nil {
+			return ctrlResult, err
+		}
+	}
+
+	// check for PKCS11 secret holding the PKCS11 Client Data
+	if slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStorePKCS11) {
```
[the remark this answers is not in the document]
vakwetu: Done

controllers/barbicanworker_controller.go
vakwetu: Done

pkg/barbican/volumes.go
```diff
@@ -138,32 +137,26 @@ func GetKollaConfigVolumeMount(serviceName string) corev1.VolumeMount {
...
-					Secret: &corev1.SecretVolumeSource{
-						DefaultMode: &config0644AccessMode,
-						SecretName:  pkcs11.CertificatesSecret,
-					},
+	return []corev1.Volume{
+		{
+			Name: P11ClientDataVolume,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					DefaultMode: &config0644AccessMode,
```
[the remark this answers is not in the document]
vakwetu: Done

controllers/barbican_controller.go
stuggi: could remove the default from SimpleCryptoBackendSecret to keep the current released behavior

stuggi: issue on my side not reading the code well in the stable branches :)

pkg/barbican/volumes.go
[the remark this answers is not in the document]
vakwetu: needed to be something like 550 because it was a directory

controllers/barbicanapi_controller.go
```diff
@@ -278,6 +278,18 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
 		"my.cnf":                             db.GetDatabaseClientConfig(tlsCfg), //(mschuppert) for now just get the default my.cnf
 	}
 
+	// Fetch the service config snippet (CustomConfigFileName) from the top
+	// level barbican controller, and add them to this service specific Secret.
+	owner := barbican.GetOwningBarbicanName(instance)
+	if owner != "" {
+		barbicanSecretName := barbican.GetOwningBarbicanName(instance) + "-config-data"
```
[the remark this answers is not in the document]
vakwetu: whoops :)

controllers/barbicankeystonelistener_controller.go
```diff
@@ -255,6 +255,19 @@ func (r *BarbicanKeystoneListenerReconciler) generateServiceConfigs(
 		barbican.CustomServiceConfigFileName: instance.Spec.CustomServiceConfig,
 		"my.cnf":                             db.GetDatabaseClientConfig(tlsCfg), //(mschuppert) for now just get the default my.cnf
 	}
+
+	// Fetch the service config snippet (CustomConfigFileName) from the top
+	// level barbican controller, and add them to this service specific Secret.
+	owner := barbican.GetOwningBarbicanName(instance)
+	if owner != "" {
+		barbicanSecretName := barbican.GetOwningBarbicanName(instance) + "-config-data"
```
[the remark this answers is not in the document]
vakwetu: whoops !

api/v1beta1/common_types.go
```diff
@@ -112,120 +112,32 @@ type BarbicanComponentTemplate struct {
...
-
-	// +kubebuilder:validation:Optional
-	// The IP address of the client connecting to the HSM (X.Y.Z.K)
-	ClientAddress string `json:"clientAddress,omitempty"`
+	// DefaultP11ClientDataPath
+	DefaultP11ClientDataPath = "/etc/hsm-client"
+)
 
+// BarbicanPKCS11Template - Includes common HSM properties
+type BarbicanPKCS11Template struct {
```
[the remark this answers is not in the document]
vakwetu: I'm not entirely clear as to what you're proposing here, but yeah, I think we want to follow the pattern we've been using from the beginning.

controllers/barbicanapi_controller.go
```diff
@@ -278,6 +278,18 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
+		barbicanSecret, _, err := secret.GetSecret(ctx, h, barbicanSecretName, instance.Namespace)
+		if err != nil {
+			return err
+		}
+		customData[barbican.CustomConfigFileName] = string(barbicanSecret.Data[barbican.CustomConfigFileName])
```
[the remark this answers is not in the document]
vakwetu: agreed.  We should do this.  But I think we should do this as a separate change.  That way we can focus on making sure that particular change doesn't break anything.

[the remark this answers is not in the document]
vakwetu: Lets open a separate Jira for this.

templates/barbican/bin/generate_p11_keys.sh
```diff
@@ -16,9 +16,12 @@
 set -xe
 
 {{- if and (index . "P11Enabled") .P11Enabled }}
-echo "Creating  MKEK label {{ .P11MKEKLabel }}"
-barbican-manage hsm check_mkek --label {{ .P11MKEKLabel }} || barbican-manage hsm gen_mkek --label {{ .P11MKEKLabel }}
 
-echo "Creating  HMAC label {{ .P11HMACLabel }}"
-barbican-manage hsm check_hmac --label {{ .P11HMACLabel }} || barbican-manage hsm gen_hmac --label {{ .P11HMACLabel }}
+mkek_label=$(grep -A 100 "\[p11_crypto_plugin\]" /etc/barbican/barbican.conf.d/01-custom.conf | grep "mkek_label" | cut -d'=' -f2|tr -d '[:space:]')
```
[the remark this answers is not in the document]
vakwetu: good one!  I didn't know we had crudini!

api/v1beta1/common_types.go
```diff
@@ -237,4 +149,7 @@ type PasswordSelector struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default="SimpleCryptoKEK"
 	SimpleCryptoKEK string `json:"simplecryptokek"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="P11CryptoLogin"
+	P11CryptoLogin string `json:"p11cryptologin"`
```
dmendiza, line 154: Nit:  Consider renaming to `PKCS11CryptoLogin` or even better `PKCS11PIN`, since this value corresponds to the PKCS#11 PIN as defined in the PKCS#11 spec.   Same reasoning as above, we should consistently use `PKCS11` everywhere instead of a mix of `PKCS11` and `P11`

api/v1beta1/barbican_webhook.go
```diff
@@ -202,6 +195,15 @@ func (r *BarbicanSpec) ValidateUpdate(old BarbicanSpec, basePath *field.Path) fi
...
 		r.BarbicanAPI.Override.Service)...)
 
+	// pkcs11 verifications
+	if slices.Contains(r.EnabledSecretStores, SecretStorePKCS11) {
+		if r.PKCS11 == nil {
+			allErrs = append(allErrs, field.Required(basePath.Child("PKCS11"),
+				"PKCS11 specification is missing, PKCS11 is required when pkcs11 is an enabled SecretStore"),
+			)
+		}
+	}
```
dmendiza, line 205: This is the same code as line 141 above, which makes me think it may be worth moving into a common place.

api/v1beta1/common_types.go
```diff
@@ -112,120 +112,32 @@ type BarbicanComponentTemplate struct {
...
-
-	// +kubebuilder:validation:Required
-	// The HSM's IPv4 address (X.Y.Z.K)
-	ServerAddress string `json:"serverAddress"`
-
-	// +kubebuilder:validation:Optional
-	// The IP address of the client connecting to the HSM (X.Y.Z.K)
-	ClientAddress string `json:"clientAddress,omitempty"`
+	// DefaultP11ClientDataPath
+	DefaultP11ClientDataPath = "/etc/hsm-client"
```
dmendiza, line 123: Nit: Consider renaming this to `DefaultPKCS11ClientDataPath`.  I think we should be consistent with naming variables, and this patch uses both `PKCS11` as in `SecretStorePKCS11` in line 120 above, and also `P11` as used here.  It may not be obvious to other folks that `P11` is short for `PKCS11` and that both will be used in the same backend.  I think it would be better and clearer to just use `PKCS11` throughout.

config/samples/pods_with_pkcs11-luna.yaml
```diff
@@ -19,31 +19,44 @@ spec:
...
+    [secretstore:pkcs11]
+    secret_store_plugin = store_crypto
+    crypto_plugin = p11_crypto
+
+    [p11_crypto_plugin]
+    plugin_name = PKCS11
+    library_path = /usr/local/luna/libs/64/libCryptoki2.so
+    token_serial_number = 12345
+    token_labels = some_token_label
+    slot_id = 123
```
dmendiza, line 37: Only one of these values is necessary.  There is an order of priority and only one of the values will be used.  It may be better for the sample to only show one.

config/samples/pods_with_pkcs11-luna.yaml
```diff
@@ -19,31 +19,44 @@ spec:
...
+    mkek_label = some_mkek_label
+    mkek_length = 123455
+    hmac_label = some_hmac_label
+    encryption_mechanism = CKM_AES_GCM
+    aes_gcm_generate_iv = true
+    hmac_key_type = CKK_GENERIC_SECRET
+    hmac_keygen_mechanism = CKM_GENERIC_SECRET_KEY_GEN
+    hmac_keywrap_mechanism = CKM_AES_KEY_WRAP_KWP
```
dmendiza, line 45: `hmac_keywrap_mechanism` is deprecated.  Use `hmac_mechanism` instead: https://opendev.org/openstack/barbican/src/branch/unmaintained/2023.1/barbican/plugin/crypto/p11_crypto.py#L87-L90

api/v1beta1/common_types.go
```diff
@@ -112,120 +112,32 @@ type BarbicanComponentTemplate struct {
...
-        // Generate IVs for CKM_AES_GCM mechanism
-	AESGCMGenerateIV bool `json:"AESGCMGenerateIV"`
-
-        // +kubebuilder:validation:Optional
-        // +kubebuilder:default=true
-        // Always set cka_sensitive
-	AlwaysSetCKASensitive bool `json:"alwaysSetCKASensitive"`
+        // +kubebuilder:validation:Required
+        // The OpenShift secret that stores the HSM client data.
+        // These will be mounted to /var/lib/config-data/hsm
```
dmendiza, line 134: Where does `/var/lib/config-data/hsm` come from?

controllers/barbican_controller.go
```diff
@@ -252,52 +252,35 @@ func (r *BarbicanReconciler) reconcileNormal(ctx context.Context, instance *barb
...
-			return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
+	if len(instance.Spec.EnabledSecretStores) == 0 || slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStoreSimpleCrypto) {
+		ctrlResult, err = r.verifySecret(ctx, helper, instance, instance.Spec.SimpleCryptoBackendSecret, []string{instance.Spec.PasswordSelectors.SimpleCryptoKEK}, &configVars)
+		if err != nil {
+			return ctrlResult, err
+		}
+	}
+
+	// check for PKCS11 secret holding the PKCS11 Crypto Login
+	if slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStorePKCS11) && instance.Spec.PKCS11 != nil {
```
dmendiza, line 270: Should we test here for `len(instance.Spec.EnabledSecretStores) == 0` as above?

controllers/barbican_controller.go
```diff
@@ -252,52 +252,35 @@ func (r *BarbicanReconciler) reconcileNormal(ctx context.Context, instance *barb
...
+		ctrlResult, err = r.verifySecret(ctx, helper, instance, instance.Spec.PKCS11.LoginSecret, []string{instance.Spec.PasswordSelectors.P11CryptoLogin}, &configVars)
+		if err != nil {
+			return ctrlResult, err
+		}
+	}
+
+	// check for PKCS11 secret holding the PKCS11 Client Data
+	if slices.Contains(instance.Spec.EnabledSecretStores, barbicanv1beta1.SecretStorePKCS11) && instance.Spec.PKCS11 != nil {
```
dmendiza, line 278: This is the same boolean test as line 270.  Consider combining into a single case.

controllers/barbican_controller.go
dmendiza, line 546: Consider using `pkcs11XXX` instead of `p11XXX` for consistency and clarity.

templates/barbican/bin/generate_p11_keys.sh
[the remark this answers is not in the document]
vakwetu: Done

controllers/barbican_controller.go
vakwetu: I don't think so.  If instance.Spec.EnabledSecretStores is [], then the first term here will fail without throwing any error -  as we'd expect.

The length check above is so that we default to simple_crypto in case EnabledSecretStores is undefined.

controllers/barbican_controller.go
vakwetu: Done

api/v1beta1/common_types.go
vakwetu: Per design discussion, we will mount the data in a standard location and use kolla to move the data to the location designated by ClientDataPath.

The path we currently use for kolla stuff is /var/lib/config-data/default.  This is just a branch close to this.

The reason for this is that if we mount the secret to ClientDataPath directly, then we end up overwriting whatever is already there.  So, if we wanted to use /usr/local/luna is the client path, we'd end up deleting everything in that directory and only having the secrets there.

config/samples/pods_with_pkcs11-luna.yaml
vakwetu: Done

config/samples/pods_with_pkcs11-luna.yaml
vakwetu: Done

config/samples/pods_with_pkcs11-luna.yaml
```diff
@@ -19,31 +19,44 @@ spec:
...
+    key_wrap_mechanism = true
```
dmendiza, line 46: This should be the name of a key wrapping mechanism, not a boolean:

`key_wrap_mechanism = CKM_AES_KEY_WRAP_WKP`

config/samples/pods_with_pkcs11-luna.yaml
dmendiza: I think this value needs to change.  The reason this option was renamed is because it is confusing.  This should be a signing algorithm used to make an hmac tag, not key wrapping.  We probably want to use the default here:

`hmac_mechanism = CKM_SHA256_HMAC`

config/samples/pods_with_pkcs11-luna.yaml
vakwetu: ack.  I noticed this when I was testing, but forgot to change it here.  Good catch.

config/samples/pods_with_pkcs11-luna.yaml
vakwetu: ack, thats what I used in testing.  Done.

api/v1beta1/common_types.go
vakwetu: Done

api/v1beta1/barbican_webhook.go
vakwetu: Done

api/v1beta1/common_types.go
vakwetu: Done

controllers/barbican_controller.go
vakwetu: Done

config/samples/pods_with_pkcs11-luna.yaml
vakwetu: Done

controllers/barbican_controller.go
```diff
@@ -555,22 +538,32 @@ func (r *BarbicanReconciler) reconcileDelete(ctx context.Context, instance *barb
...
+	caBundleSecretNameField  = ".spec.tls.caBundleSecretName"
+	tlsAPIInternalField      = ".spec.tls.api.internal.secretName"
+	tlsAPIPublicField        = ".spec.tls.api.public.secretName"
+	p11LoginSecretField      = ".spec.pkcs11.loginSecret"
+	p11ClientDataSecretField = ".spec.pkcs11.clientDataSecret"
 )
 
 var (
-	commonWatchFields = []string{
+	workerWatchFields = []string{
```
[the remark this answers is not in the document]
stuggi: yes, lets follow up on this. here in barbican it is, right now, not an issue, since all controller render the global config, due to the link used for the templates directory. if the components would just consume the global config ,rendered by the main controller, and this one does not index/watch properly the input secret, the component controllers would reconcile, but not getting an updated rendered global config of the updated secret since the main controller did not reconcile. so something to keep in mind when removing the link as mentioned bellow, since there could be a situation where the user updated the config secret, but still using the old config since it has not rendered again.

controllers/barbicanapi_controller.go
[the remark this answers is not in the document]
stuggi: ack sounds good

vakwetu:
Just merged all the changes into a single commit to trigger new kuttl test run now that https://github.com/openshift/release/pull/60903 has merged.

stuggi (approved):
/lgtm lets address things we discussed in this PR and not yet in, in follow ups

PR #203, opened 2025-02-07
Fix generate_pkcs11_keys.sh to handle quotes

vakwetu:
Right now, generate_pkcs11_keys.sh will read from the config and take the value of the mkek and hmac labels as-is, without removing any quotes.  So, if the config file has quotes around the value, like "foo", the key "foo" (with quotes) will be created.

Of course, when barbican-api comes up, it strips the quotes and starts looking for key foo(and cannot find it).

This fixes the problem by stripping quotes from the value in the script.


dmendiza:
lgtm

dmendiza (approved):

PR #197, opened 2025-01-03
Add configurable API Timeout

Deydra71:
...to allow configure timeouts for HAProxy and Apache.

Jira: https://issues.redhat.com/browse/OSPRH-10964

PR #207, opened 2025-02-14
Set APITimeout minimum validation

Deydra71:
...to prevent user from setting a negative value. Minimum value of 10 follows common implementation in other operators.


Deydra71:
@afaranha 
Basic functional test was included in the original patch here - https://github.com/openstack-k8s-operators/barbican-operator/pull/197

PR #228, opened 2025-03-19
Refactor the volume definition in the deployments

vakwetu:
This is a refactoring of how secrets are mounted in barbican
to more closely align with the design that is implied by our CRD.

Up to now, many of the fields in the CRD were incompletely
implemented.  An explanation of how the secrets are expected to be
mounted is provided in some developer documentation in docs/volumes.md.

In particular, in this patch, we
- implmented the defined but incorrectly implmented
  CustomServiceConfigSecrets.  This includes verifying the existence
  of the secrets and adding a watch for any changes.
- unlinking the template directories.  This results in templates
  not being duplicated in multiple secrets - and eliminates the need
  to redundantly generate the templates in the child CRs.
- refactor the volume/mount definition code to make it more readable.
- eliminate logging.conf as we do not actually use this file currently
  and if we did, we do not currently override any defaults.
- implement the defined but unimplemented DefaultConfigOverwrite
- Add functional tests to confirm that all the relevant config shows
  up in all the right secrets and that these secrets are mounted.
- Fix the value for host_href which was incorrectly being set to a
  constant value.  It should be set to "" (empty string) to force
  barbican to extract the href from the request.
- Added code to retrieve DefaultConfigOverwrite from the parent CR.

Jira: [OSPRH-13924](https://issues.redhat.com//browse/OSPRH-13924)


controllers/barbicanapi_controller.go
```diff
@@ -936,7 +867,41 @@ func (r *BarbicanAPIReconciler) reconcileNormal(ctx context.Context, instance *b
...
+		apis := &barbicanv1beta1.BarbicanAPIList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(namespace),
+		}
+		if err := r.Client.List(context.Background(), apis, listOpts...); err != nil {
+			Log.Error(err, "Unable to retrieve barbicanapi CRs %v")
+			return nil
+		}
+		// Watch for changes to any CustomServiceConfigSecrets
+		for _, cr := range apis.Items {
```
mauricioharley, line 887: Just wondering if we can ever hit some O(N^2) problem with this nested loop.  Is it something we should care about?

controllers/barbicankeystonelistener_controller.go
```diff
@@ -651,7 +632,41 @@ func (r *BarbicanKeystoneListenerReconciler) reconcileNormal(ctx context.Context
...
+		listeners := &barbicanv1beta1.BarbicanKeystoneListenerList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(namespace),
+		}
+		if err := r.Client.List(context.Background(), listeners, listOpts...); err != nil {
+			Log.Error(err, "Unable to retrieve barbicankeystonelistener CRs %v")
+			return nil
+		}
+		// Watch for changes to any CustomServiceConfigSecrets
+		for _, cr := range listeners.Items {
```
mauricioharley, line 652: Same eventual performance issue as above.  Applicable?

controllers/barbicanapi_controller.go
```diff
@@ -270,129 +267,54 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
-	if err != nil {
-		return err
+	customSecrets := ""
+	for _, secretName := range instance.Spec.CustomServiceConfigSecrets {
+		secret, _, err := secret.GetSecret(ctx, h, secretName, instance.Namespace)
+		if err != nil {
+			return err
+		}
+		for _, data := range secret.Data {
+			customSecrets += string(data) + "\n"
```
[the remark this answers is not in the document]
vakwetu: Thats a good question.  I'm following cinder's interpretation of what these secrets are supposed to look like - and this is essentially lifted from their implementation.  Basically, what is expected here are snippets that would be read by oslo.config.  The difference between what is here and what is in CustomServiceConfig is that these snippets can include secrets.  See https://github.com/openstack-k8s-operators/cinder-operator/blob/58c8786a75d5e6dd5acd872566e495b9f53258bb/docs/user-guide/commonalities.md?plain=1#L258 for a more detailed explanation.

controllers/barbicanapi_controller.go
vakwetu: I don't think so.  The outer loop is going to be on the order of 1 - and in most cases, likely to be 1.  Also, the number of secrets on the inner loop is likely to be small - if not zero.

controllers/barbicankeystonelistener_controller.go
vakwetu: ditto above.  Likely not to be an issue.

pkg/barbican/dbsync.go
```diff
@@ -16,53 +16,18 @@ const (
...
-			MountPath: "/etc/barbican/barbican.conf.d",
-			ReadOnly:  true,
-		},
-		{
-			Name:      "config-data",
-			MountPath: "/var/lib/kolla/config_files/config.json",
-			SubPath:   "barbican-dbsync-config.json",
-			ReadOnly:  true,
-		},
+		GetKollaConfigVolumeMount(instance.Name + "-dbsync"),
```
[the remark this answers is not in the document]
vakwetu: No.  This is still correct.  The instance name here is "barbican".

GetKollaConfigVolumeMount(service_name) uses the service_name to identify the correct json file from the default secret.  

SubPath:   serviceName + "-config.json"

pkg/barbicankeystonelistener/deployment.go
```diff
@@ -27,35 +27,12 @@ func Deployment(
...
-		barbican.GetKollaConfigVolumeMount(instance.Name),
-		barbican.GetLogVolumeMount(),
-	}
-
-	// Add the CA bundle
-	if instance.Spec.TLS.CaBundleSecretName != "" {
-		keystoneListenerVolumes = append(keystoneListenerVolumes, instance.Spec.TLS.CreateVolume())
-		keystoneListenerVolumeMounts = append(keystoneListenerVolumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
-	}
+	keystoneListenerVolumes, keystoneListenerVolumeMounts := GetVolumesAndMounts(instance)
```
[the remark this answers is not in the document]
vakwetu: Opted to change the function names to make everything super clear.
Now they are GetAPIVolumesAndMounts(), GetWorkerVolumesAndMounts() and GetListenerVolumesAndMounts()

I decided not to include an error parameter because right now, no error is thrown.  If I did have one, I'd have to catch and handle the error in the calling function, which in turn would require changing the signature of the calling function (which doesn't yet throw an error) .. etc.  Not sure if we want to add all that extra (and at this point, unused) code.

pkg/barbicankeystonelistener/volumes.go
```diff
@@ -0,0 +1,34 @@
+package barbicankeystonelistener
+
+import (
+	barbicanv1beta1 "github.com/openstack-k8s-operators/barbican-operator/api/v1beta1"
+	barbican "github.com/openstack-k8s-operators/barbican-operator/pkg/barbican"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// GetVolumesAndMounts - return the volumes and mounts for a BarbicanKeystoneListener deployment
+func GetVolumesAndMounts(instance *barbicanv1beta1.BarbicanKeystoneListener) ([]corev1.Volume, []corev1.VolumeMount) {
```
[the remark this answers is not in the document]
vakwetu: See above - chose to rename.

pkg/barbicanworker/volumes.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"slices"
+
+	barbicanv1beta1 "github.com/openstack-k8s-operators/barbican-operator/api/v1beta1"
+	barbican "github.com/openstack-k8s-operators/barbican-operator/pkg/barbican"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// GetVolumesAndMounts - return the volumes and mounts for a BarbicanWorker deployment
+func GetVolumesAndMounts(instance *barbicanv1beta1.BarbicanWorker) ([]corev1.Volume, []corev1.VolumeMount) {
```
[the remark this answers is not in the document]
vakwetu: See above - chose to rename.

controllers/barbicanapi_controller.go
```diff
@@ -270,129 +267,54 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
-			barbican.DatabaseName,
-		),
-		"KeystoneAuthURL":  keystoneInternalURL,
-		"ServicePassword":  string(ospSecret.Data[instance.Spec.PasswordSelectors.Service]),
-		"ServiceUser":      instance.Spec.ServiceUser,
-		"ServiceURL":       "https://barbican.openstack.svc:9311",
-		"TransportURL":     string(transportURLSecret.Data["transport_url"]),
-		"LogFile":          fmt.Sprintf("%s%s.log", barbican.BarbicanLogPath, instance.Name),
-		"EnableSecureRBAC": instance.Spec.EnableSecureRBAC,
+		"ServiceURL": "https://barbican.openstack.svc:9311",
```
stuggi, line 310: which svc url is this? we create two services `barbican-internal` and `barbican-public`. there seems to be no `barbican`, also
* this should probably not be hard coded to the `openstack` namespace.  
* could it be non `https`? probably not recommended, but what happens when internal tls is disabled?

controllers/barbicanapi_controller.go
```diff
@@ -936,7 +867,41 @@ func (r *BarbicanAPIReconciler) reconcileNormal(ctx context.Context, instance *b
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *BarbicanAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *BarbicanAPIReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	Log := r.GetLogger(ctx)
+	secretFn := func(_ context.Context, o client.Object) []reconcile.Request {
```
stuggi, line 872: not tested, but I think we could just index it like we do with the passwordSecret and such. 

```
	// index customServiceConfigSecrets
	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &barbicanv1beta1.BarbicanAPI{}, customServiceConfigSecretsField, func(rawObj client.Object) []string {
		// Extract the customServiceConfigSecrets names from the spec, if one is provided
		cr := rawObj.(*barbicanv1beta1.BarbicanAPI)
		
		return cr.Spec.CustomServiceConfigSecrets
	}); err != nil {
		return err
	}
```

controllers/barbicankeystonelistener_controller.go
```diff
@@ -651,7 +632,41 @@ func (r *BarbicanKeystoneListenerReconciler) reconcileNormal(ctx context.Context
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *BarbicanKeystoneListenerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *BarbicanKeystoneListenerReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	Log := r.GetLogger(ctx)
+	secretFn := func(_ context.Context, o client.Object) []reconcile.Request {
```
stuggi, line 637: see above

controllers/barbicanworker_controller.go
```diff
@@ -666,7 +630,41 @@ func (r *BarbicanWorkerReconciler) reconcileNormal(ctx context.Context, instance
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *BarbicanWorkerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *BarbicanWorkerReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	Log := r.GetLogger(ctx)
+	secretFn := func(_ context.Context, o client.Object) []reconcile.Request {
```
stuggi, line 635: see above

controllers/barbicankeystonelistener_controller.go
vakwetu: ditto as above

controllers/barbicanworker_controller.go
vakwetu: ditto as above

controllers/barbicanapi_controller.go
vakwetu: investigating this ..

vakwetu: Basically, this is the value that barbican puts in the href for a secret.  ie.  if you do an "openstack secret list", you see a value like "https://barbican.openstack.svc:9311/v1/secrets/sdfhhdfsk".  It turns out that doing "openstack secret get https://... " strips out the href part anyways, so it doesn't matter. 

The right thing to do though - and its what I've done in the latest version - is to set that parameter to "" (empty string).  This will cause barbican to populate the value from the HTTP request -so we'll get the right hostname and http/https.

Note - we do have to set it to "".  Not doing that will set the default value which is "http://localhost/..."

xek (approved):
LGTM!

tests/kuttl/tests/barbican_tls/01-assert.yaml
```diff
@@ -150,7 +153,7 @@ spec:
         - name: config-data
           secret:
             defaultMode: 420
-            secretName: barbican-api-config-data
```
frenzyfriday, line 153: This seems to break barbican kuttl tests: https://logserver.rdoproject.org/7f3/rdoproject.org/7f32adc940f1472bb2c1446088c79a26/controller/ci-framework-data/logs/ci_script_006_run_barbican.log

PR #225, opened 2025-03-14
Fix redundant Barbican pod name suffixes

Deydra71:
Jira: [OSPRH-14708](https://issues.redhat.com/browse/OSPRH-14708)

PR #231, opened 2025-03-27
Gracefully clean up old deployments with duplicate names

Deydra71:
After [#225](https://github.com/openstack-k8s-operators/barbican-operator/pull/225), old Barbican deployments with duplicate name suffixes (e.g., barbican-api-api) are no longer automatically removed.

This patch introduces a cleanup mechanism that attempts to gracefully scale down and delete old deployments once the new deployments are ready.

### DIFF
--- a/controllers/barbican_controller.go
+++ b/controllers/barbican_controller.go
@@ -50,6 +50,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	"golang.org/x/exp/maps"
+	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -1151,4 +1152,27 @@ func (r *BarbicanReconciler) ensureDB(
 	instance.Status.DatabaseHostname = db.GetDatabaseHostname()
 	instance.Status.Conditions.MarkTrue(condition.DBReadyCondition, condition.DBReadyMessage)
 	return db, ctrlResult, nil
+}
+
+func cleanupOldDeployment(ctx context.Context, c client.Client, namespace string, oldName string) error {
+	var oldDeployment appsv1.Deployment
+	err := c.Get(ctx, types.NamespacedName{Name: oldName, Namespace: namespace}, &oldDeployment)
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	if oldDeployment.Spec.Replicas == nil || *oldDeployment.Spec.Replicas > 0 {
+		replicas := int32(0)
+		oldDeployment.Spec.Replicas = &replicas
+		if updateErr := c.Update(ctx, &oldDeployment); updateErr != nil {
+			return fmt.Errorf("failed to scale down old deployment '%s': %w", oldName, updateErr)
+		}
+		return nil
+	}
+
+	if delErr := c.Delete(ctx, &oldDeployment); delErr != nil {
+		return fmt.Errorf("failed to delete old deployment '%s': %w", oldName, delErr)
+	}
+
+	return nil
 }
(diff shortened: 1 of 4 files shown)


Deydra71:
/retest

quayi.io timeout

controllers/barbican_controller.go
```diff
@@ -1152,3 +1153,26 @@ func (r *BarbicanReconciler) ensureDB(
...
+		if updateErr := c.Update(ctx, &oldDeployment); updateErr != nil {
+			return fmt.Errorf("failed to scale down old deployment '%s': %w", oldName, updateErr)
+		}
+		return nil
+	}
+
+	if delErr := c.Delete(ctx, &oldDeployment); delErr != nil {
+		return fmt.Errorf("failed to delete old deployment '%s': %w", oldName, delErr)
+	}
+
```
stuggi, line 1176: wouldn't it be enough to only run the delete and ignore the `NotFound`, which would be a single api call, instead of 3 ?
```
	deployment := &appsv1.Deployment{
		ObjectMeta: metav1.ObjectMeta{
			Name:      oldName,
			Namespace: namespace,
		},
	}

	err := c.Delete(ctx, deployment)
	if err != nil && !k8s_errors.IsNotFound(err) {
		return fmt.Errorf("Error cleanup old deployment %s: %w", oldName, err)
	}
```

controllers/barbicanapi_controller.go
```diff
@@ -924,6 +924,15 @@ func (r *BarbicanAPIReconciler) reconcileNormal(ctx context.Context, instance *b
 	}
```
stuggi, line 924: If we call the cleanup here when the new deployment got its ReadyCondition and just delete the old deployment, ignoring NotFound error, wouldn't that be graceful as the new one is already up??

Related, see https://github.com/openstack-k8s-operators/barbican-operator/pull/235

controllers/barbican_controller.go
stuggi: in a later release, when we are comfortable that users were on a release with the cleanup, could remove it again.

controllers/barbicanapi_controller.go
Deydra71: I see. So, after the #235 is merged, we can use the IsReady() to check if the new deployment is ready, and after that we can directly delete the old deployment and ignoring the `NotFOund `error.

controllers/barbicanworker_controller.go
```diff
@@ -655,6 +655,15 @@ func (r *BarbicanWorkerReconciler) reconcileNormal(ctx context.Context, instance
 	}
 	// create Deployment - end
 
+	if depl.GetDeployment().Status.ReadyReplicas > 0 {
+		oldDepName := fmt.Sprintf("%s-api", instance.Name)
```
Deydra71, line 659: oldDepName := fmt.Sprintf("%s-worker", instance.Name)

controllers/barbican_controller.go
```diff
@@ -1185,3 +1186,18 @@ func (r *BarbicanReconciler) ensureDB(
...
 }
+
+func cleanupOldDeployment(ctx context.Context, c client.Client, namespace, oldName string) error {
+	oldDeployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      oldName,
+			Namespace: namespace,
+		},
+	}
+
```
stuggi, line 1197: one thing we probably should do here is to validate that the corresponding barbican CR is the owner of the old deployments before delete it. in theory there could be barbican deployments from different owner CRs. similar to https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/galera.go#L55 . just a safety mechanism to not delete something the one being reconciled does not own.

stuggi (commented):
could also add a test in https://github.com/openstack-k8s-operators/barbican-operator/blob/main/tests/functional/barbican_controller_test.go ? like when adding whatever test deployment with the old name and the owner ref of the CR it gets deleted.

tests/functional/barbican_controller_test.go
```diff
@@ -1334,6 +1337,87 @@ var _ = Describe("Barbican controller", func() {
 
 	})
 
+	When("a Barbican CR is created and an old deployment with proper owner reference exists", func() {
```
Deydra71, line 1340: The new added functional test is failing for now. I'm not sure is the cleanupOldDeployment function is being called properly to cleanup the fake old deployment, or if it's valid to create fake barbican deployment in the functional test runtime like this.

Will get back to it later.

Deydra71:
The bug fix has been manually verified - https://issues.redhat.com/browse/OSPRH-17555

PR #234, opened 2025-04-01
Run Barbican services without root

fmount:
As done for several service operators, this patch removes root from the `Barbican` services.
`SecurityContexts` are created to address requirements related to different services.

Jira: https://issues.redhat.com/browse/OSPRH-15340

### DIFF
--- a/pkg/barbican/const.go
+++ b/pkg/barbican/const.go
@@ -65,4 +65,8 @@ const (
 	PKCS11ClientDataVolume = "pkcs11-client-data"
 	// PKCS11DataVolume is the mount point used for PKCS11 client Data
 	PKCS11ClientDataMountPoint = "/var/lib/config-data/hsm"
+	// BarbicanUID
+	BarbicanUID int64 = 42403
+	// BarbicanGID
+	BarbicanGID int64 = 42403
 )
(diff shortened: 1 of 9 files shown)


fmount:
Let's see CI in action first

mauricioharley (commented):
Cool!

fmount:
awesome, this should be good now!

pkg/barbican/const.go
```diff
@@ -61,4 +61,8 @@ const (
 	PKCS11ClientDataVolume = "pkcs11-client-data"
 	// PKCS11DataVolume is the mount point used for PKCS11 client Data
 	PKCS11ClientDataMountPoint = "/var/lib/config-data/hsm"
+	// BarbicanUID - based on https://github.com/openstack/kolla/blob/master/kolla/common/users.py
```
stuggi, line 64: should we reference to tcib instead? https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/uid_gid_manage.sh#L37

stuggi (commented):
just nit, otherwise lgtm

pkg/barbican/const.go
fmount: oh yeah, it makes sense, updating the patch with the link you suggested! Thank you!

fmount:
> awesome, this should be good now!

Rebased and good to go.

fmount:
@stuggi @abays you ok to land this?

abays (approved):
/lgtm

PR #248, opened 2025-05-12
[WIP] AppCred service operator support

Deydra71:
This PR adds end-to-end support for consuming Keystone ApplicationCredentials (AC) in the Barbican operator, enabling BarbicanAPI, BarbicanWorker and KeystoneListener to fall back to AC-based auth when available.

AC watches:
*`AddACWatches` registers watches on ApplicationCredential CRs and their corresponding Secrets (that are created by openstack-operator)
* Controllers enqueue a reconcile for any CR named `ac-<service>` and its `-secret` Secret

Reconcile logic:
* On each reconcile the controller looks for an ApplicationCredential CR
* If the CR is missing, it logs and will continue using password authentication
* If the CR exists but its Secret isn’t ready yet, it logs that it’s waiting and requeues

Once the Secret appears, it:
* Reads the AC_ID and AC_SECRET fields and injects them into the service’s configuration so the pod will authenticate with application credentials instead of a password.
* Computes a hash of the Secret’s contents and stores that in the status, changes trigger new hash computation and a rolling update of the Deployment whenever the credentials rotate.

SetupWithManager changes:
Adds field indexer for AC CR and AC Secret. 

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/567

### DIFF
--- a/controllers/barbican_controller.go
+++ b/controllers/barbican_controller.go
@@ -94,6 +94,7 @@ func (r *BarbicanReconciler) GetLogger(ctx context.Context) logr.Logger {
 //+kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneapis,verbs=get;list;watch;
 //+kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneservices,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneendpoints,verbs=get;list;watch;create;update;patch;delete;
+//+kubebuilder:rbac:groups=keystone.openstack.org,resources=applicationcredentials,verbs=get;list;watch
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete;
@@ -108,6 +109,7 @@ func (r *BarbicanReconciler) GetLogger(ctx context.Context) logr.Logger {
 
 // service account, role, rolebinding
 //+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
 //+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=roles,verbs=get;list;watch;create;update;patch
 //+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=get;list;watch;create;update;patch
 //+kubebuilder:rbac:groups="security.openshift.io",resourceNames=anyuid,resources=securitycontextconstraints,verbs=use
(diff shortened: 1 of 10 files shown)


controllers/barbican_controller.go
```diff
@@ -661,6 +663,25 @@ func (r *BarbicanReconciler) generateServiceConfig(
...
+		// fetch AC Secret
+		secret := &corev1.Secret{}
+		secName := types.NamespacedName{Namespace: ac.Namespace, Name: ac.Status.SecretName}
+		if err := r.Client.Get(ctx, secName, secret); err == nil {
+			// switch to application credentials auth
+			templateParameters["UseApplicationCredentials"] = true
+			templateParameters["ACID"] = string(secret.Data["AC_ID"])
+			templateParameters["ACSecret"] = string(secret.Data["AC_SECRET"])
+			Log.Info("Using ApplicationCredentials auth")
+		}
```
vakwetu, line 680: How do we handle the case where we are unable to retrieve the ACSecret?  Looks like we simply continue and use password auth instead.  But maybe we should requeue?  Or at the very least log the error?

controllers/barbicanapi_controller.go
```diff
@@ -612,6 +612,36 @@ func (r *BarbicanAPIReconciler) reconcileNormal(ctx context.Context, instance *b
...
+			helper.GetClient(),
+			10*time.Second,
+		)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if res.RequeueAfter > 0 {
+			return res, nil
+		}
+		configVars["secret-"+secretName] = env.SetValue(hashAC)
```
vakwetu, line 642: Why do we need this logic?  It appears to be different from all the other secret checks above.

controllers/barbicanapi_controller.go
```diff
@@ -612,6 +612,36 @@ func (r *BarbicanAPIReconciler) reconcileNormal(ctx context.Context, instance *b
 
 	Log.Info(fmt.Sprintf("[API] Got secrets '%s'", instance.Name))
 
+	// check for  ApplicationCredential
```
vakwetu, line 615: Should we be checking for both the admin secret and also the application credential (if it exists)?
Seems to me that if you are using appCreds - then you no longer need to watch the admin secret - and no longer need to reconcile if it changes.

Maybe the logic here should be more like:
check if app cred exists 
if app_cred_exists {
    get app_cred_secret()
    if ! exists requeue
    verify app_cred_secret
} else {
   verify_admin_secre
}

Feels like this is long enough to warrant a new method verify_service_credentials()

controllers/barbicankeystonelistener_controller.go
```diff
@@ -401,6 +402,36 @@ func (r *BarbicanKeystoneListenerReconciler) reconcileNormal(ctx context.Context
...
+		)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if res.RequeueAfter > 0 {
+			return res, nil
+		}
+		configVars["secret-"+secretName] = env.SetValue(hashAC)
+	}
+
```
vakwetu, line 434: Same comments as above, but actually, we should think about whether we need this logic at all.  I don't think that the barbican-keystone-listener actually talks to keystone or uses the keystone creds.  In which case, maybe we don't need to verify these parameters at all.

controllers/barbicanworker_controller.go
```diff
@@ -399,6 +400,35 @@ func (r *BarbicanWorkerReconciler) reconcileNormal(ctx context.Context, instance
...
+			10*time.Second,
+		)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if res.RequeueAfter > 0 {
+			return res, nil
+		}
+		configVars["secret-"+secretName] = env.SetValue(hashAC)
+	}
```
vakwetu, line 431: Same comments as above, but actually, we should think about whether we need this logic at all. I don't think that the barbican-worker actually talks to keystone or uses the keystone creds. In which case, maybe we don't need to verify these parameters at all.

controllers/barbicanapi_controller.go
vakwetu: Also, the above check needs to be in barbican_controller too.  See https://github.com/openstack-k8s-operators/barbican-operator/blob/main/controllers/barbican_controller.go#L257

Deydra71: I added the `verifyApplicationCredentials()` func that implements this, see the comment below

PR #251, opened 2025-05-19
Accommodating multiple HSM vendors

mauricioharley:
This patch updates the functional tests structure to accommodate multiple HSM vendors, and also inserts the tests for Eviden Trustway Proteccio HSMs.

Jira:  [OSPRH-14751](https://issues.redhat.com//browse/OSPRH-14751)

### DIFF
--- a/tests/functional/barbican_controller_test.go
+++ b/tests/functional/barbican_controller_test.go
@@ -859,12 +859,12 @@ var _ = Describe("Barbican controller", func() {
 		})
 	})
 
-	When("A Barbican with pkcs11 plugin is created", func() {
+	When("A Barbican with PKCS11 plugin is created - Luna model", func() {
 		BeforeEach(func() {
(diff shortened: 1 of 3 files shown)


vakwetu (commented):
I'm in two minds as to the need for this PR.

We did a good enough job the first time round, such that no new code needed to be added to the barbican-operator to accomodate the proteccio HSM.  That means that the code path followed by both the proteccio and luna HSMs is the same - its just basically the code path of a generic pkcs11 object.

We can see this because the difference between the luna test and the proteccio test is just a file path and different contents of the secrets involved, and the test steps added are essentially the same as those for luna.

Given that the test steps are the same, and the code paths are the same - I'm not sure it makes sense to add any further test code. 

If we really do want to add something that is HSM specific, then I think we need to extract the test code to a function so that the steps are not duplicated in the test code.  Otherwise if we do end up modifying the processing of a generic pkcs11 object, then we will have multiple places to fix up.

Apply the DRY principle here : don't repeat yourself.

vakwetu (changes requested):
see comment ..

PR #288, opened 2025-09-26
Modernize Go code with latest style conventions

fmount:
Modernize using `gopls modernize tool` to update:

- Replace interface{} with any type alias (Go 1.18+)
- Update range loops to use idiomatic range syntax
- Replace fmt.Sprintf with fmt.Appendf where appropriate

These changes improve code readability and align with current Go best practices.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/controllers/barbicanapi_controller.go
+++ b/controllers/barbicanapi_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"maps"
 	"slices"
 	"time"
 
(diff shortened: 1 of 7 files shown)


controllers/barbicanapi_controller.go
```diff
@@ -312,15 +313,11 @@ func (r *BarbicanAPIReconciler) generateServiceConfigs(
...
-		}
+		maps.Copy(customData, ownerInstance.Spec.DefaultConfigOverwrite)
 	}
 
-	for key, data := range instance.Spec.DefaultConfigOverwrite {
-		// can we merge here?
-		customData[key] = data
-	}
+	// can we merge here?
+	maps.Copy(customData, instance.Spec.DefaultConfigOverwrite)
```
abays, line 320: The answer is yes, so I wonder if we should just remove this?

PR #349, opened 2026-03-08
Register missing field indexes for Secret watcher in Barbican controller

mauricioharley:
The Barbican controller's findObjectsForSrc uses FieldSelector queries on 7 fields to map Secret events to Barbican CRs that need reconciliation. However, only authAppCredSecretField had a corresponding IndexField registration in SetupWithManager(). The remaining 6 fields (passwordSecretField, simpleCryptoBackendSecretField, caBundleSecretNameField, pkcs11LoginSecretField,
pkcs11ClientDataSecretField, customServiceConfigSecretsField) were missing, causing "Index with name field:... does not exist" errors on every Secret event in the namespace.

Two of these fields (caBundleSecretNameField,
customServiceConfigSecretsField) do not exist on the parent Barbican type — they live in the child specs (BarbicanAPISpec, etc.). Their indexes are registered returning nil to satisfy the FieldSelector contract without producing false matches.

### DIFF
--- a/internal/controller/barbican_controller.go
+++ b/internal/controller/barbican_controller.go
@@ -680,6 +680,50 @@ func (r *BarbicanReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
+	// Index passwordSecretField
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &barbicanv1beta1.Barbican{}, passwordSecretField, func(rawObj client.Object) []string {
+		cr := rawObj.(*barbicanv1beta1.Barbican)
+		if cr.Spec.Secret == "" {
+			return nil
+		}
+		return []string{cr.Spec.Secret}
+	}); err != nil {
+		return err
+	}
+
+	// Index simpleCryptoBackendSecretField
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &barbicanv1beta1.Barbican{}, simpleCryptoBackendSecretField, func(rawObj client.Object) []string {
+		cr := rawObj.(*barbicanv1beta1.Barbican)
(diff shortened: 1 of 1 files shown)


Deydra71 (commented):
I digged deeper into the CIX issue and this fix and found out that the https://github.com/openstack-k8s-operators/barbican-operator/pull/335 introduced two bugs -  
* missing IndexField registrations for parent-level secret fields (`passwordSecretField`, `simpleCryptoBackendSecretField`, `pkcs11LoginSecretField`, `pkcs11ClientDataSecretField`)
* two child-only fields (`caBundleSecretNameField`, `customServiceConfigSecretsField`) included in the parent's findObjectsForSrc loop, but  these fields don't exist on the Barbican spec, they live in BarbicanComponentTemplate (child specs only) and are already correctly indexed and watched by each child controller.

I don't think that introducing dummy `IndexField` returning nil for `caBundleSecretNameField` and `customServiceConfigSecretsField` on the parent Barbican type is a good way, mroe like a workaround that masks the second bug. 

Looking at nova example - https://github.com/openstack-k8s-operators/nova-operator/blob/main/internal/controller/nova_controller.go#L2311 has `passwordSecretField` and `authAppCredSecretField` in its watch list, and registers IndexField only for those two. The are indexed only in child controllers. This is consistent in other operators as well.

I tested a fix that adds `IndexField` registrations for the 4 parent-level fields that were missing and removed `caBundleSecretNameField` and `customServiceConfigSecretsField` from the parent's `findObjectsForSrc` loop -https://github.com/Deydra71/barbican-operator/tree/testing-fix-cix-1238 and oeprator logged no errors, and it correctly reconciles if the watched fields change/content is updated.

Deydra71 (approved):

PR #356, opened 2026-04-01
Add application credential finalizer management

Deydra71:
Jira: [OSPRH-27509](https://redhat.atlassian.net/browse/OSPRH-27509)

Application Credential dev-doc: https://github.com/openstack-k8s-operators/dev-docs/blob/main/application_credentials.md

* Tracks the active AC secret name in `Status.ApplicationCredentialSecret`
* Add `openstack.org/barbican-ac-consumer` finalizer to the AC secret after service config is rendered
* On AC rotation, move the finalizer from the old secret to the new one
* On CR deletion, remove the consumer finalizer from the AC secret before cleaning up the CR

This ensures that the keystone-operator cannot revoke a rotated AC secret while Barbican is still consuming it.

```
2026-04-28T11:47:49Z	INFO	Controllers.Barbican	Added consumer finalizer	{"controller": "barbican", "controllerGroup": "barbican.openstack.org", "controllerKind": "Barbican", "Barbican": {"name":"barbican","namespace":"openstack"}, "namespace": "openstack", "name": "barbican", "reconcileID": "f36c0bd9-d99c-4c90-8337-7e5017c93a55", "object": "ac-barbican-37683-secret", "finalizer": "openstack.org/barbican-ac-consumer"}
2026-04-28T11:47:49Z	INFO	Controllers.Barbican	Removed consumer finalizer	{"controller": "barbican", "controllerGroup": "barbican.openstack.org", "controllerKind": "Barbican", "Barbican": {"name":"barbican","namespace":"openstack"}, "namespace": "openstack", "name": "barbican", "reconcileID": "f36c0bd9-d99c-4c90-8337-7e5017c93a55", "object": "ac-barbican-d6d05-secret", "finalizer": "openstack.org/barbican-ac-consumer"}

```

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/685

Assisted-by: Claude Opus 4.6 [noreply@anthropic.com](mailto:noreply@anthropic.com)


internal/controller/barbican_controller.go
```diff
@@ -611,13 +616,78 @@ func (r *BarbicanReconciler) reconcileDelete(ctx context.Context, instance *barb
...
+		return nil
+	}
+
+	var newObj, oldObj client.Object
+
+	if newSecretName != "" {
+		secret := &corev1.Secret{}
+		key := types.NamespacedName{Name: newSecretName, Namespace: instance.Namespace}
+		if err := h.GetClient().Get(ctx, key, secret); err != nil {
+			return fmt.Errorf("failed to get new AC secret %s: %w", newSecretName, err)
```
mauricioharley: This wraps every error the same way, including `NotFound`. In practice, if the openstack-operator points `Spec.Auth.ApplicationCredentialSecret` to a new secret before the keystone-operator has actually created it, you'll get a stream of errors in the logs for something that's perfectly normal and resolves on its own.

What about treating `NotFound` as a softer case?

```go
if k8s_errors.IsNotFound(err) {
    Log.Info("AC secret not yet available, will retry", "secret", newSecretName)
    return nil
}
return fmt.Errorf("failed to get new AC secret %s: %w", newSecretName, err)
```

That way the logs stay clean during the normal window where keystone hasn't caught up yet.

test/functional/barbican_controller_test.go
```diff
@@ -1803,6 +1803,144 @@ var _ = Describe("Barbican controller", func() {
...
+					ContainElement(barbican.ACConsumerFinalizer))
+			}, timeout, interval).Should(Succeed())
+
+			// Status should reflect the new secret
+			Eventually(func(g Gomega) {
+				b := GetBarbican(barbicanTest.Instance)
+				g.Expect(b.Status.ApplicationCredentialSecret).To(Equal(newACSecretName))
+			}, timeout, interval).Should(Succeed())
+		})
+	})
```
mauricioharley, line 1962: Nice set of tests here - covers the add, the status tracking, and the rotation move. One thing I'm missing though: what happens when the Barbican CR itself gets deleted? The `reconcileDelete` logic removes the consumer finalizer from the AC secret, but there's no test exercising that path.

Something along the lines of:

```go
It("should remove the consumer finalizer from AC secret on Barbican CR deletion", func() {
    // Wait for the consumer finalizer to appear on the AC secret
    // ...
    // Delete the Barbican CR
    // ...
    // Verify the AC secret no longer carries the barbican-ac-consumer finalizer
    // ...
})
```

Would be good to have that covered so regressions don't sneak in.

mauricioharley (changes requested):
Please check the inline comments.

internal/controller/barbican_controller.go
Deydra71: I'm trying to think about a scenario when this situation would actually happen...

In keystone-operator -  k8s Secret is created first, then `Status.SecretName` is set, then the AC CR is marked Ready all in the same reconcile + deferred status.

openstack-operator reads the secretn ame from `Status.SecretName` only after AC is marked as Ready, then it is written to the service CR https://github.com/openstack-k8s-operators/openstack-operator/blob/main/internal/openstack/barbican.go#L107 from where the service reads it. openstack-operator never propagates a secret name that doesn't already exist, because it gates on `acCR.IsReady()`

Deydra71:
recheck

internal/controller/barbican_controller.go
```diff
@@ -389,6 +389,23 @@ func (r *BarbicanReconciler) reconcileNormal(ctx context.Context, instance *barb
...
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.ServiceConfigReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.ServiceConfigReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
+		}
+	}
+	instance.Status.ApplicationCredentialSecret = instance.Spec.Auth.ApplicationCredentialSecret
```
vakwetu: Here is a suggestion from claude:

Status update pattern violation (line 407)
  
  Setting instance.Status.ApplicationCredentialSecret here is problematic because it happens after the finalizer is successfully added but before deployment is ready.

  The issue:
  - If any later reconciliation step fails (deployment creation, network setup, readiness check), this in-memory status update is lost
  - This creates an inconsistent state where:
    - The finalizer WAS added to the new secret (persisted)
    - The status still shows the old secret name (update lost)
    
  Example failure scenario:
  1. Line 407: Set status.ApplicationCredentialSecret = "secret-v2" (in memory)
  2. Line 450+: Start creating Deployment with new config
  3. Deployment creation fails (quota, network error, etc.)
  4. Reconciliation returns error, status update is never persisted
  5. Next reconcile sees: spec="secret-v2", status="secret-v1", finalizer already on secret-v2

  Recommended fix:
  Move this status update to the end of reconcileNormal(), just before setting ReadyCondition to true, after confirming the deployment is actually ready:

  // Only update status after deployment is confirmed ready
  if deployment.Status.ReadyReplicas > 0 {
      instance.Status.ApplicationCredentialSecret = instance.Spec.Auth.ApplicationCredentialSecret
      instance.Status.Conditions.MarkTrue(condition.ReadyCondition, ...)
  }

  This ensures status truthfully reflects: "We successfully protected this secret AND deployed pods using it."

test/functional/barbican_controller_test.go
```diff
@@ -1974,6 +1974,164 @@ var _ = Describe("Barbican controller", func() {
...
+					Namespace: barbicanTest.Instance.Namespace,
+					Name:      acSecretName,
+				})
+				g.Expect(secret.Finalizers).To(
+					ContainElement(barbican.ACConsumerFinalizer))
+			}, timeout, interval).Should(Succeed())
+
+			th.DeleteInstance(GetBarbican(barbicanTest.Instance))
+
+			secret := th.GetSecret(types.NamespacedName{
```
vakwetu: Flaky deletion test (lines 2123-2131)
    - Checks finalizer removal immediately after DeleteInstance() without waiting
    - Will fail randomly when deletion reconciliation is slow
    - Fix: Wrap finalizer check in Eventually() block

go.mod
```diff
@@ -10,14 +10,14 @@ require (
 	github.com/onsi/gomega v1.41.0
 	github.com/openstack-k8s-operators/barbican-operator/api v0.0.0-00010101000000-000000000000
 	github.com/openstack-k8s-operators/infra-operator/apis v0.6.1-0.20260513130700-78e1a15a8289
-	github.com/openstack-k8s-operators/keystone-operator/api v0.6.1-0.20260513130126-175a0958de92
+	github.com/openstack-k8s-operators/keystone-operator/api v0.6.1-0.20260520090027-4d7b7a01c0bf
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.6.1-0.20260518125357-72bdd580c587
-	github.com/openstack-k8s-operators/lib-common/modules/storage v0.6.1-0.20260515134210-2e2a0d06648c
-	github.com/openstack-k8s-operators/lib-common/modules/test v0.6.1-0.20260515134210-2e2a0d06648c
-	github.com/openstack-k8s-operators/mariadb-operator/api v0.6.1-0.20260513143847-4b70b899997a
+	github.com/openstack-k8s-operators/lib-common/modules/storage v0.6.1-0.20260506154724-30a976ba8ef0
```
vakwetu: Some of these are being downgraded.  Is this correct?

vakwetu (changes requested):
looks good in  general. Just a couple points that claude picked up on.

internal/controller/barbican_controller.go
Deydra71: Status is not lost on error, because `helper.PatchInstance` runs always in defer. 

But the status update touches edge case situation described in watcher - https://github.com/openstack-k8s-operators/watcher-operator/pull/369#discussion_r3234900678 and moving the status update only after deployment is ready is not sufficient, because the manage finalizer function relies on comparing secret in status vs in spec. So, the correct pattern is to split:

* Early: Add finalizer to the new (spec) secret only -- passing `""` as the old secret to `ManageACSecretFinalizer`, so nothing is removed.
* Late (after all sub-services are Ready): Call `RemoveACSecretConsumerFinalizer` on the old secret + update `status.ApplicationCredentialSecret`.

For the non-rotation case, status is set immediately since there's no old secret to protect.

go.mod
Deydra71: I think this is because of manual bump I did to keystone and lib-common so we don't need to wait for automatic bump. I will rebase on main now, but in any way, it's not a problem because automated CI bump would correct it.

fmount:
Seems in line with all other operators that we just merged. @Deydra71 I assume we can land this one, it looks good, but let me know if we want to wait the green light from other reviewers.

vakwetu (approved):
LGTM

Deydra71:
Had to rebase

fmount (approved):
/lgtm

PR #380, opened 2026-05-13
Fix defaultConfigOverwrite mounting to /etc/barbican/

Deydra71:
Jira: [OSPRH-25781](https://redhat.atlassian.net/browse/OSPRH-25781)

The `defaultConfigOverwrite` field is documented to place files in /etc/barbican/, but barbican-operator was mounting them into /etc/barbican/barbican.conf.d/ This prevented overriding files that upstream barbican hardcodes to /etc/barbican/ (e.g. barbican-api-paste.ini) and was inconsistent with how other operators (nova, keystone, etc.) handle this field.

Fixed the bug by adding per-file SubPath volume mounts from the existing config-data-custom Secret into /etc/barbican/<key>. SubPath mounts are used instead of a directory mount to avoid shadowing existing files in /etc/barbican/.

For backwards compatibility, each defaultConfigOverwrite key is mounted at BOTH locations:
  1. /etc/barbican/<key> (correct path per CRD)
  2. /etc/barbican/barbican.conf.d/<key> (old path)

The dual mount preserves existing customer workarounds where e.g. policy.yaml was referenced via `customServiceConfig` as:
```
  [oslo_policy]
  policy_file = /etc/barbican/barbican.conf.d/policy.yaml
```


Deydra71:
/test barbican-operator-build-deploy-kuttl

seems like a unrelated timeout

Deydra71:
Needed to update kuttl assert to check correct number of volumeMounts

vyzigold:
I tested the PR similar to what we have in our [draft documentation](https://redhat.atlassian.net/wiki/spaces/CLOUDOPS/pages/389779771/Manually+enable+audit+logs#Barbican) for audit logging (providing a custom api-paste.ini with the audit middleware enabled). I can confirm this PR works for customizing the api-paste.ini and is what we need for the audit logs. :+1:

vakwetu (approved):
LGTM

PR #383, opened 2026-05-21
Add ExtraMounts support for custom volume injection

mauricioharley:
Implements ExtraMounts to allow operators to inject custom volumes and mounts into Barbican components (API, Worker, KeystoneListener, DbSync), following the pattern established by other OpenStack operators (heat, cinder, manila, etc.).

This enables use cases such as mounting custom HTTPD configuration files (e.g. SSL/TLS hardening) via ConfigMaps or Secrets without modifying the operator code.

Includes functional tests validating ExtraMounts propagation to sub-CRs and volume presence in deployments, along with sample HTTPD override files with PQC-aware TLS guidance.

Closes: [OSPRH-30350](https://redhat.atlassian.net/browse/OSPRH-30350)


config/samples/httpd-overrides/README.md
```diff
@@ -0,0 +1,139 @@
...
+  barbican:
+    enabled: true
+    template:
+      extraMounts:
+      - extraVol:
+        - extraVolType: httpd-overrides
+          propagation:
+          - BarbicanAPI
+          mounts:
+          - mountPath: /etc/httpd/conf_custom
```
fmount, line 81: @mauricioharley to make this work and not conflict with the existing mounts, I think we need an `IncludeOptional` directive in `httpd.conf`. In storage we currently have it within the `vhost` [1] (which is included in the main `httpd.conf` that is something we could also dedup).

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/templates/cinder/config/10-cinder_wsgi.conf#L40

internal/barbican/dbsync.go
```diff
@@ -25,6 +25,21 @@ func DbSyncJob(instance *barbicanv1beta1.Barbican, labels map[string]string, ann
 		dbSyncMounts = append(dbSyncMounts, instance.Spec.BarbicanAPI.TLS.CreateVolumeMounts(nil)...)
 	}
 
+	// ExtraMounts
+	propagation := []storage.PropagationType{Barbican}
```
fmount: Do we need extraMounts propagation in dbsync? Usually this `job` only needs db connection, so we recently make it minimal and only mount `00-config.conf`, which is the global file where the db connect string is rendered [1].

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/internal/cinder/job.go#L45

mauricioharley:
@fmount, thank you for the review.  Mind taking a look again?

internal/barbicanapi/volumes.go
```diff
@@ -62,5 +62,20 @@ func GetAPIVolumesAndMounts(instance *barbicanv1beta1.BarbicanAPI, overwriteKeys
 		apiVolumeMounts = append(apiVolumeMounts, barbican.GetHSMVolumeMounts()...)
 	}
 
+	// ExtraMounts
+	propagation := []storage.PropagationType{barbican.BarbicanAPI, barbican.Barbican}
+	for _, exv := range instance.Spec.ExtraMounts {
+		for _, vol := range exv.Propagate(propagation) {
```
fmount: wondering if this block, that is common for api, keystonelistener and worker can be dedup with a common helper in `internal/barbican/volumes.go` like the following:

```golang
  func GetExtraVolumes(extraMounts []barbicanv1beta1.BarbicanExtraVolMounts, svc []storage.PropagationType) ([]corev1.Volume, []corev1.VolumeMount) {
        var volumes []corev1.Volume
        var mounts []corev1.VolumeMount
        for _, exv := range extraMounts {
                for _, vol := range exv.Propagate(svc) {
                        for _, v := range vol.Volumes {
                                volumeSource, _ := v.ToCoreVolumeSource()
                                volumes = append(volumes, corev1.Volume{
                                        Name:         v.Name,
                                        VolumeSource: *volumeSource,
                                })
                        }
                        mounts = append(mounts, vol.Mounts...)
                }
        }
        return volumes, mounts
  }
```

fmount (commented):
@mauricioharley overall the change looks good. Thanks for that. I only have a suggestion to deduplicate the code replicated for api, keystonelistener, worker, otherwise lgtm.

fmount (approved):
/lgtm

PR #391, opened 2026-06-23
Remove eval and xtrace from PKCS11 key generation script

stuggi:
Remove unnecessary eval from crudini variable assignments — direct command substitution works the same without shell interpretation of the result. Remove -x from set -xe to prevent echoing of HSM PIN values to stdout/logs. Quote variable expansions in barbican-manage arguments.

Jira: [OSPRH-31726](https://redhat.atlassian.net/browse/OSPRH-31726)


Deydra71 (approved):
/lgtm

PR #392, opened 2026-06-25
Set service_type in [keystone_authtoken] for access rule validation

Deydra71:
Without service_type configured, keystonemiddleware cannot validate application credentials with custom access rules, causing HTTP 401 for end users.

Closes: [OSPRH-22365](https://redhat.atlassian.net/browse/OSPRH-22365)


fmount (approved):
/lgtm

PR #399, opened 2026-06-30
Fix client IP logging in Barbican API VirtualHost

fmount:
The `VirtualHost` `CustomLog` directive was overriding the server-level dual-log setup, causing all requests to be logged with the OpenShift router IP instead of the real client IP from `X-Forwarded-For`.

Add `SetEnvIf` and dual `CustomLog` lines inside the `VirtualHost` block to match the `nova-operator` pattern.

Jira: [OSPRH-32126](https://redhat.atlassian.net/browse/OSPRH-32126)

### DIFF
--- a/templates/barbican/config/10-barbican_wsgi_main.conf
+++ b/templates/barbican/config/10-barbican_wsgi_main.conf
@@ -17,9 +17,11 @@
   </Directory>
 
   ## Logging
-  ErrorLog "/var/log/barbican/error.log"
+  ErrorLog /dev/stdout
   ServerSignature Off
-  CustomLog "/var/log/barbican/access.log" combined env=!forwarded
+  SetEnvIf X-Forwarded-For "^.*\..*\..*\..*" forwarded
+  CustomLog /dev/stdout combined env=!forwarded
+  CustomLog /dev/stdout proxy env=forwarded
 
 {{- if $vhost.TLS }}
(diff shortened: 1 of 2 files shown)


templates/barbican/config/10-barbican_wsgi_main.conf
```diff
@@ -19,7 +19,9 @@
   ## Logging
   ErrorLog "/var/log/barbican/error.log"
   ServerSignature Off
+  SetEnvIf X-Forwarded-For "^.*\..*\..*\..*" forwarded
   CustomLog "/var/log/barbican/access.log" combined env=!forwarded
+  CustomLog "/var/log/barbican/access.log" forwarded env=forwarded
```
stuggi: wondering if logging to access.log and above to error.log is ok, or should it go to stdout like we do in the others?

fmount: Yes, we should go to /dev/stdout like others. Let me follow up here and we can make it consistent as part of the same change.

fmount: ok @stuggi done, now barbican is consistent w/ the other operators' pattern and we can stream logs via `oc logs -f` because previously we had `access.log` within the container.

stuggi (approved):
/lgtm thanks a lot @fmount for fixing this!

PR #407, opened 2026-07-24
Remove unnecessary pods RBAC permissions

stuggi:
Operators create Deployments/StatefulSets/Jobs to run workload pods but never manage Pod objects directly (get/list/watch/create/update/patch/delete). Drop the unused "pods" resource from the kubebuilder RBAC markers and the workload-facing ClusterRole rbacRules, and regenerate config/rbac/role.yaml accordingly.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,7 +8,6 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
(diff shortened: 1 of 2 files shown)


stuggi:
recheck

abays (approved):
/lgtm

PR #418, opened 2026-07-28
[18-stable] Remove unnecessary pods RBAC permissions

stuggi:
The workload rbacRules and the matching kubebuilder RBAC marker in the controller granted the workload service account full CRUD (create/delete/get/list/patch/update/watch) on core Pods, but that service account never reads or writes Pod objects directly. Remove the unused grants.

The kubebuilder RBAC marker for pods (get;list) is kept: lib-common's VerifyNetworkStatusFromAnnotation, called from this controller to verify NetworkAttachments, lists Pods using the controller-manager's own client, so the manager's ClusterRole (config/rbac/role.yaml) still needs get;list on pods. Regenerate config/rbac/role.yaml.

Includes b3f8fc57596cd4a336ee432b5a250fb991ecd6ab

(cherry picked from commit 31d93a863945b238c68535230219f54ccfe61293)

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,7 +8,6 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
   - secrets
   - services
   verbs:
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

PR #422, opened 2026-08-03
Debug logging off by default

Deydra71:
Operator templates shipped debug = true and httpd LogLevel debug, so every deployment was verbose by default (noise + possible sensitive data in logs). 

Defaults `debug = false` and `LogLevel warn` (as in e.g. octavia’s httpd and other services). Pods get non-debug logging unless the user overrides via custom config.


Deydra71:
/test barbican-operator-build-deploy-kuttl

timed out

